PR #38, opened 2020-10-26
Rolling

jomjol:


MiguelAngelLV:
Maybe, a interesting option for web interface access could be check a mqtt topic before enter in deep sleep, so, if "not_sleep" is enabled not enter en deep sleep.

So, you only need wait to the next cycle.

PR #84, opened 2021-01-03
Removed "(in seconds)" from path-configs

savikko:


jomjol:
Incorporated in latest rolling (2021-01-18).
Thanks for the info and correction.

PR #142, opened 2021-03-09
fix: typo at config page -> mqtt section

IPSAttain:
the wrong sample for the mqtt broker irritated me during the installation


jomjol:
Incorporated in v6.2.2 

Thanks a lot!

PR #263, opened 2021-07-11
Rolling

Zwer2k:
Implemented the first version of the GPIO handler. For me the version runs very stable. I have added a list of changes in the readme.


jomjol:
Looking like a lot of work! Great and thanks for that improvement.

Never the less, brings me to a bit of thinking, how to take over, as I'm about to release a major change to finally enable the option for more than one number on one picture - like a dual meter.
Therefore I need to change quite some in the internal data handling - almost done, but also on the html interface, not yet published.

It might take some time to bring this together. Maye we can have a chat about this offline?

PR #287, opened 2021-07-24
removed ExtendedResolution which was displayed always on top

mdax82:
on top of index_configure.html under "Config.ini" a "ExtendedResolution" was displayed which had no function and its still in the Analog settings. Hence removed the line which seem to work still afterwards.


jomjol:
Included in the last rolling / master

PR #380, opened 2021-10-30
fix missing inverted commas

stan23:
... in class description


jomjol:
Updated in latest rolling - Thanks for this fix.

PR #384, opened 2021-11-01
improvements on the usage of SmartLed

stan23:
- create a global instance of SmartLed when the flash is used for the first time, instead of using a local instance which is destructed automatically at the end of `GpioHandler::flashLightEnable()` and may lead to inaccessible memory when `SmartLed::copyRmtHalfBlock()` is called from interrupt context later
- remove duplicate files


stan23:
reference ad discussion: https://forum.iobroker.net/post/696742

jomjol:
Does it still work in your environment. I have implemented your global LED in my code and it constantly reboots after the first run with an error of the camera. That might be related to GPIO problems.
Now I have two uncomfortable states:
1) use local led variable and initialize it every time I use it --> working only up to two LEDs
2) use global led --> does not work at all.

I will stay with 1) for the time beeing. Any ideas?

stan23:
Yes it still works for me. And I also went back to your code (no changes) and that crashes with more than 2 LEDs.
I'm fine if you keep it as it is, since it works for you and you are the maintainer :)

I'll stick to my changed version for the time being, no problem.

jomjol:
Sure - but a version, that is not limited to 2 would be very benefical. The changes are so simple, that don't think, that I made an error. It must be some kind of hardware error.
I'll keep on searching. For the time beeing I stay with my code, as the other it not running at all currently in my system?!

stan23:
I'll soon setup a second ESP32-CAM for my gasmeter and will check both your version and my branch.
Let's see if it behaves differently for me and if yes, whether I can find out anything why it doesn't work.

jomjol:
I just tested the globale code on another ESP32CAM and it is working! Looks like a hardware problem with my camera module :-(.
I just uploaded the global code to the rolling version. You can switch from global to local definition easily by a compiler option in the header file.
Maybe you want to test it as well. If it will run stable on my 4 ESP32 productive systems for some days I will push it to the master.

jomjol:
In the newest version, this problem has been tackeled and now more than 2 LEDs can be used.

PR #439, opened 2021-12-03
Time based flow rate limiting

### LINKED ISSUE #438: Feature Request: Limit rate of change based on time since last value
Right now the rate of change (MaxRateValue) is simply based on the last cycle, whenever that was. There are scenarios where this can lead to the value no longer updating, e.g. when there was no valid detection of the last digit in several cycles until the point where MaxRateValue is exceeded.

If the Max Rate was relative to time (e.g. per minute), then I could set a reasonable maximum rate of change for my household (e.g. 0.1kw per minute). Then even when the time between updates is longer, I still have a reasonable upper bound of allowed change.

040medien:
Fixes #438  
Take time since last valid value into account for flow rate limiting


040medien:
@jomjol I've been thinking a bit more whether we should even store / send the current rate value in case of a rate error. Happy to change this if you prefer.

PR #440, opened 2021-12-03
don't store flow rate in case of errors

040medien:
Don't store the flow rate and also don't send it by MQTT if it exceeds the max value

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
@@ -658,9 +658,9 @@ bool ClassFlowPostProcessing::doFlow(string zwtime)
 
             double difference = difftime(imagetime, NUMBERS[j]->lastvalue);      // in Sekunden
             difference /= 60;                                                    // in Minuten
-            NUMBERS[j]->FlowRateAct = (NUMBERS[j]->Value - NUMBERS[j]->PreValue) / difference;
+            double flowrate = (NUMBERS[j]->Value - NUMBERS[j]->PreValue) / difference;
             
-            if (NUMBERS[j]->useMaxRateValue && (abs(NUMBERS[j]->FlowRateAct) > NUMBERS[j]->MaxRateValue))
+            if (NUMBERS[j]->useMaxRateValue && (abs(flowrate) > NUMBERS[j]->MaxRateValue))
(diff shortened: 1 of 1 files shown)


jomjol:
The question is: what to send, if the reading is inconsistently.

Current status:
- send  the old value, set the rate to "0"
- send an error message

If I got your idea right:
- send only the error message and nothing else

I like this idea, but to do so, I need to modify the MQTT part as well or I did not understand the changes in the code. I think about it.

040medien:
Actually my first PR changed this behavior since it started to send the
erroneous rate. So in my second PR I reverted that.

By the way, the new logic fixed my unsteady meter readings - before that it
often had jumped ahead of the actual reading or stopped entirely because it
had missed a few cycles, especially when the meter was moving faster. Now
my meter graph is super smooth.

jomjol ***@***.***> schrieb am Sa., 4. Dez. 2021, 08:06:

spcqike:
today, i finally set up an esp-cam. i still need to tinker around with leds/light, as i have some reflections.

i also noticed, that, if the value / rate is to high, it sends this value via MQTT. 
this leads to strange graphics.
it looks like, that this just happens when the new value is to high. if its smaller, the rate is reported as "0"

grafik
grafik

grafik

040medien:
Yep, that's what I saw as well. @jomjol are considering merging this PR or do you want to refactor the approach? I think the current state of affairs is a bit inconsistent since the rate gets sent even if no valid value gets sent.

jomjol:
In the newest `rolling` I changed the behavious. Values are only transfered via MQTT, if there is no error. 
This handling has a long history and would need some code cleaning, but currently I don't have time for this.

jomjol:
This has been implemented in the newest rolling - but not been testet deeply.

spcqike:
I just updated to the latest rolling firmware. Will report back how it goes. 

Just one additional question. 
Is there an overview of all numbers / fonts that are taught into the system? 
i know, my illumination is not good as I have some reflections, but somehow my “4” is recognized as “0”. 
DE6127EB-AC18-4198-8189-FF226B40311D

i will print an enclosure within the next weeks, this should fix my light problem.

jomjol:
You can find the trained number here: https://github.com/jomjol/neural-network-digital-hybrid-readout/tree/main/data_raw_all

PR #466, opened 2021-12-20
DeepSleep

error105:
Hi, mayby u will think about deepsleep option ? Normally nobody needs keep espcam always on, and with 2000mah LIION and deepslepp it can works e lot of time if we only wake up it few times in day :)


jomjol:
Deep sleep mode is not really working for this device. Even with all optimzations you will not go below 5-8mA, due to the electrical circuet and the used components.

You would need to change some of the components (power converter, transistor) or have another board.

error105:
I know that i need mak changes on board, but still i neet that option in firmware to use it :(

jomjol:
Okay, as I have no hardware to debug this and it quite some effort, I'm not going to impolent this for the free version. If you have interest in an commercial solution, please contact me through my email (see main page).

PR #770, opened 2022-04-13
Clean up hardcoded IP addresses from gethost.js

### LINKED ISSUE #769: IP addresses in HTML files
Thanks for developing this project, it works very well!

To isolate this device from the rest of our network, I run it on a separate WiFi network and then access it via some tunnels. When I do so, some of the UI fails to load. While investigating this I noticed that the HTML files stored on the SD card contain some static IP addresses, such as
https://github.com/jomjol/AI-on-the-edge-device/blob/a020fce2d794f4a56f7bb65b266982c7554d08b7/sd-card/html/edit_alignment.html#L107

I wonder if those should not be replaced by
```js
basepath = window.location.protocol + "//" + window.location.host
```
or something along those lines, so that the UI works regardless of how the UI is served?

wetneb:
Closes #769.

This lets me access the UI via an SSH tunnel as my device is connected to a protected network.

### DIFF
--- a/sd-card/html/gethost.js
+++ b/sd-card/html/gethost.js
@@ -5,25 +5,9 @@ function gethost_Version(){
 
 
 function getbasepath(){
-    var host = window.location.hostname;
-    if (((host == "127.0.0.1") || (host == "localhost") || (host == "")) 
-//       && ((window.location.port == "80") || (window.location.port == ""))
-       )
-    
-    {
-//        host = "http://192.168.2.219";          // jomjol interner test
-//        host = "http://192.168.178.46";          // jomjol interner test
-        host = "http://192.168.178.34";          // jomjol interner Real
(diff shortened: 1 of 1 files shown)


wetneb:
Note that there are a lot more static IPs in the rest of the HTML files, which would be worth cleaning up I think, but this seemed to be the critical thing to fix in my case.

jomjol:
If I merge this, my debugging does not work anymore. Is there any change to detect, wheter you are using a tunnel or if it is in debugging mode?

wetneb:
You could set those IP addresses in a localStorage field which would be read by `gethost.js`, perhaps?
https://developer.mozilla.org/de/docs/Web/API/Window/localStorage

jomjol:
This is needed for local debugging (not very elegant, but working :-) ).
Therefore cannot be commited.

PR #774, opened 2022-04-15
InfluxDB integration

### LINKED ISSUE #534: Directly write to InfluxDB
Hej,

I am about to switch from my own water meter readout setup ["watermeter project"](https://github.com/bolausson/watermeter) to this amazing project.

I am wondering if there are any plans to e.g. write directly to a InfluxDB server using e.g. [influxdb-cpp](https://github.com/orca-zhang/influxdb-cpp) and make the connection configurable via the Web-Interface.

My C/C++ skills are "read only" - so as much as I would love to contribute this, I can't :(

_By the way, I have generated a annotated library [picture of a digit <-> value it shows] to train my neural network. It contains chopped of numbers, distorted numbers etc which have manually been assigned a value and has a size of 924 MB - let me know if this might be useful for you. My best model can be found [here](https://github.com/bolausson/watermeter/tree/main/training/model)_

Cheers,
Bjoern

wetneb:
Closes #534.

We already had a library `esp_http_client` as a dependency, which I used to write the integration, so this does not add any new dependency and has a negligible impact on firmware size.

@bolausson for now this only logs the meter value and nothing else, we can consider adding more. For now I am mostly requesting feedback on the overall structure.


bolausson:
Thanks a lot for implementing this!

To be honest, I can't figure out the "measurement" structure  which will be written to InfluxDB (lack of C knowledge I guess) :-)
Could you provide a mockup of the measurement you are writing to InfluxDB?

I would propose something like the below mockup:
`[
	{
	"measurement": "AI-on-the-edge-device",
	"time": datetime.date.strftime(currenttime, '%Y-%m-%dT%X.%z'),
	"tags": {
		"Location": "Kitchen|Bathroom|...",
		"Type": "Warm water|Cold water|Gas|...",
		"Unit": "m³|kWh|..."
		},
	"fields": {
		"value": float(value),
                "raw_value": float(raw_value),
                "score": float(value_recognition_score)
		}
	}
]`

Since this little device is very versatile, it makes sense to let the user create his own tag:value pairs via the WEB-UI (if not the request will pour in to add the Tag "color", "size"... you get the idea).

wetneb:
For now the measurement is simply:
```
measurement main/value=389.2389 12398726847
````
There are no tags and a single value.

I could add a field in the configuration UI to specify a static list of tags (or can you think of any tags which would vary over time?)

jomjol:
Merged to the rolling for the first testing - will be part of master in a later release

bolausson:
> For now the measurement is simply:
> 
> ```
> measurement main/value=389.2389 12398726847
> ```
> 
> There are no tags and a single value.
> 
> I could add a field in the configuration UI to specify a static list of tags (or can you think of any tags which would vary over time?)

Mhhh, no, I guess once you placed the device on a metering device, things will not change (unless you use it for something else). So if you can create a list of key:value tags during setup, I don't see why you would want to change them later - unless you have a typo or similar...

Issue #410, opened 2021-11-20
Javascript errors on setup pages


jensb:
This is a really cool evaluation of the original water-meter system. I was looking forward to trying it out, and flashing and Wifi connection worked without a flaw.
However, when I get to page two of the setup process, no image is shown. The browser console tells me this:

```
readconfigparam.js:14 Uncaught TypeError: Cannot read property 'split' of undefined
    at ParseConfig (readconfigparam.js:14)
    at init (edit_reference.html:275)
    at edit_reference.html:450
ParseConfig @ readconfigparam.js:14
init @ edit_reference.html:275
(anonymous) @ edit_reference.html:450
edit_reference.html:130 Uncaught TypeError: Cannot read property 'Alignment' of undefined
    at loadRawImage (edit_reference.html:130)
    at HTMLInputElement.onclick (edit_reference.html:42)
loadRawImage @ edit_reference.html:130
onclick @ edit_reference.html:42 
```

This happens both on Chromium and Brave (Firefox based) browsers. 

No image is shown and I cannot take one.

Any ideas what causes this?

jackyyc:
I also encountered the same mistake as you.

jensb:
I managed to avoid these errors by patching html/readconfigcommon.js, but the base problem is that I can only reach my ESP32 via a SSH tunnel (different Wifi network) and the setup constructs absolute URLs with the "real" IP of the ESP.

Can we have relative URLs please?

jomjol:
Can send me the patch, than I will merge it into the code. 

Regarding the absolute URLs. I agree, that the html code is not very good. But honestly, 
1) html is not my core competence and I'm happy, that it is working as it is.
2) I don't have ressources and time to work onthis. Currently I'm the only one who is actively developing this.

So I'm absolutly happy for any support and will merge any improvements as soon as possible :-) .
Looking forward to your patch.

Best regards,
   jomjol

jensb:
This gets the setup procedure running without JS errors even when using a SSH tunnel to access the ESP32:
(My smart devices use their own Wifi provided by a Raspberry Pi to separate them from the home Wifi and also provide better signal coverage also in the basement. Thus I need a SSH tunnel.)

It assumes your local test system at 192.168.178.79 is accessed using port 80.  If the port is different to 80, the code now assumes that a tunnel is being used and creates basepaths with correct hostname and port.

Since your project currently has no licence: this patch is hereby placed in the public domain. 

```
[...]

PR #866, opened 2022-07-19
re-add port check

### LINKED ISSUE #865: Access web-ui via SSH-Tunnel
Hi,

I'm accessing port 80 of my ESP32 via a SSH-Tunnel and get some errors, because using e.g. localhost:3123 makes the UI-JavaScript add your Test-Systems-IP to the URLs.
There was already a fix merged in #410 and I guess it was commented out by accident.

meisterlampe:
Remove comment to re-enable access via ssh-tunnels ( see #410  )
closes #865

### DIFF
--- a/sd-card/html/gethost.js
+++ b/sd-card/html/gethost.js
@@ -7,7 +7,7 @@ function gethost_Version(){
 function getbasepath(){
     var host = window.location.hostname;
     if (((host == "127.0.0.1") || (host == "localhost") || (host == "")) 
-//       && ((window.location.port == "80") || (window.location.port == ""))
+       && ((window.location.port == "80") || (window.location.port == ""))
        )
     
     {


jomjol:
The problem is, that I use this localport for my local debugging of the html page. If on my PC "localhost" is detected, I switch the communication to a IP adress, that is one of my ESP32s. This way I can debug the html page on a normal PC.

There are other ways to handle this, but currently I don't want to change it.

sd-card/html/gethost.js
```diff
@@ -7,7 +7,7 @@ function gethost_Version(){
 function getbasepath(){
     var host = window.location.hostname;
     if (((host == "127.0.0.1") || (host == "localhost") || (host == "")) 
-//       && ((window.location.port == "80") || (window.location.port == ""))
+       && ((window.location.port == "80") || (window.location.port == ""))
```
cristianmitran, line 10: what's the change here? :)

jomjol:
No change anymore!

meisterlampe:
The change is, that the port check is included again, which is currently commented. This enables users to access the UI via a ssh-tunnel, where the adress of the esp would be e.g. http://localhost:3123. On the other hand you should be able to access the UI on your development machine using http://localhost as before (that is the check for port 80).

PR #918, opened 2022-08-18
Update FeatureRequest.md

ppisljar:


jomjol:
Low power consumption not dominates by ESP 32, but the camera module and more quality power converter on ESP32CAM device, therefore no further attempt towards deep sleep or comparable useful

Issue #919, opened 2022-08-18
False post-process with analog + digital (class100/cont)


haverland:
On water meters with analog and digitals on the new models the lowest digital will be false calculated if the highest analog >=9.5 and the lowest digital ist > x.5

Rate too high - Read: 73.98163 - Pre: 72.95581
in this case the 

2.7 for the 2nd digit and 
9.8 for the highest analog

jomjol:
Implemented in v11.1.0

PR #920, opened 2022-08-18
fix for #919 the prev is int, so <9.0 instead of <9.5

haverland:


jomjol:
This is accepted in the rolling version and will go to the main with the next update.

PR #931, opened 2022-08-21
added favicon and adjusted window title to show hostname first

caco3:


caco3:
The icon is explicitly based on public work, so there are no legal issue (see [readme](images/icon/readme.md)).

PR #952, opened 2022-08-28
Implements Counter Clockwise Analog Dials, adds CPU Temp , RSSI to status

### LINKED ISSUE #361: Analog needles that increase counter clockwise
What a great project!!

I have a gas meter with two needles that increase as expected clockwise while it has two analog needles that increase counter clockwise. Is there a way to configure the model to detect these analog needles correctly?

gas_meter

Note the 1,000,000 and 10,000 cu ft indicators increase counter clockwise.

Thanks in advance for your help!

-- Ken

(I suspect that I will also need to reduce the intensity of the reflection over the top of the 100,000 and 10,000 indicators)

### LINKED ISSUE #941: [Enhancement] Show RSSI in Info page
AFAIK, currently the only way to see the RSSI is through MQTT.

I think it would be helpful to show it somewhere, eg. on the Info page.

adellafave:
Added a checkbox to ROI for Analog Digits to enable "Counter Clockwise Rotation". Implemented using "10 - value from CNN" to identify the correct value of digit. Verified functionality on my meter. Reading the correct values.

I also added CPU Temperature and RSSI to the Status box on the main webpage of the gui. I have my ESP mounted outdoors in front of my electric meter, so both of these will be critical. Added CPU Temp to MQTT.

closes #941 
closes #361


jomjol:
Really cool, that you made it happen. I would like to test it, but have no time currenlty. Might take a view days, but will defentelty go to the next version

adellafave:
No problem, sounds good. I love the project. Have one on my water meter too. Thanks for all the hard work.

PR #958, opened 2022-08-30
Improve ota page

caco3:
Made the OTA page look a bit more compact so part 4 is not hidden underneath the iframe:
grafik


caco3:
Do I also need to update the html.zip file or will do it?

There will come several other WebUI PR, so it might be worth to just wait :)

jomjol:
Thanks! I will do the html.zip with the next updates.

P.S.: please commit firstly to the rolling (I changed it here already).

caco3:
Thanks!

And sorry about the rolling. It picks master as default and I forgot to change it.
maybe you could set rolling as default branch?

PR #976, opened 2022-09-03
 Older Firmware versions (11.2.0 and older) do not support css files, therefore we need to keep it integrated for now! 

caco3:


jomjol:
I don't think, we need this. It's recommended to update firmware and html at one time. You should not use an old firmware with a new html.

What do you think?

caco3:
The problem is if you do a downgrade, eg. from rolling to last release.

I would merge this for now and then in half a year or so, switch to the CSS file usage again.

PR #985, opened 2022-09-06
Testcase for #950

haverland:
- only test case, added optional model type (default Digital100) for testing older models


haverland:
sorry, false branch

PR #989, opened 2022-09-07
Update web UI

caco3:
* moved `jomjols` name from the pages title to the info page, added links there to the Github page
* modified OTA page to make separation between firmware/html and CNN better. renamed `html/` to `Web Interface`, as the user does not care how it is build
* added link to releases page


caco3:
@jomjol If you are not happy with the moving of your name from the pages title to the info page, let me know and I will revert it!

The OTA page now has a link which points to the [release](https://github.com/jomjol/AI-on-the-edge-device/releases) page, so users can easier check for updates.

I would suggest to delete the `firmware` folder. The output of a build process should not be part of the repo. Instead we should attach those files to the releases. I have done this as an example for the [11.2.0](https://github.com/jomjol/AI-on-the-edge-device/releases/tag/v11.2.0) release:
grafik

Also for the rolling releases we could create releases. Simply tick the checkbox to mark it specially:
grafik

jomjol:
Thanks for you effort. Following remarks:
1) I'm fine with moving my name to the info page.
2) I don't intent to work with releases for the rolling. This is really for experts and should not be used otherwise
3) I will stay with the firmware in the repo, cause I use this here to update all my internal devices
4) Regarding the naming of the html.zip I was thinking about to change the update slightly:
  * merge the firmware.bin and html.zip into one file (e.g. update.zip) and upload alway both.
  * this way we are sure to have a fitting html and firmware
  * user does not need to think, which to update and why two.
  
  What do you think about (4) ?
But I would like to stay with the firmware inside my

caco3:
>     2. I don't intent to work with releases for the rolling. This is really for experts and should not be used otherwise

That makes sense.

>     3. I will stay with the firmware in the repo, cause I use this here to update all my internal devices

Hmm, ok, its just not common practice.
But do you agree to add the files (additionally) as assets to a release?
This gives the users a better overview, together with the (nice!) changelog you always write.

>     4. Regarding the naming of the html.zip I was thinking about to change the update slightly:
>      * merge the firmware.bin and html.zip into one file (e.g. update.zip) and upload alway both.
>      * this way we are sure to have a fitting html and firmware
>      * user does not need to think, which to update and why two.

That sounds like a good idea!
Some thoughts:
 - Instead of `update.zip`, we should name it with the version, eg. `AI-on-the-edge-device-11.2.0.zip`. Else you could run into a mess when you download it multiple times.
 - For development, it still should be possible to only update the firmware or the web interface. Do you think that's somehow possible?
 - It would be helpful to show an update progress on the web interface. In general the upload step and flashing/unzipping steps could be combined. The status then would iterate through `IDLE`, `UPLOADING`, `UNPACKING`, `FLASHING_FIRMWARE`, `UPDATING_WEB_INTERFACE`, `DONE`.

jomjol:
sounds good and sure is it okay to add the additional files into the release.
Only with the version I struggle a bit, because the zip file is created automatically and has no info about the version (yet?).

Anyway this will take some time to implement, as I'm busy with some other stuff during September.

caco3:
I strongly would suggest that we let the releases build automatically, e. using Github actions.
Do you already use some such actions?
Else I could have a look if could set it up.
Then the zip file would get named automatically, usually based on the release tag name.

jomjol:
Good idea and no I don`t have any acitivities running. If you would take care, I would appreciate.

caco3:
> Good idea and no I don`t have any acitivities running. If you would take care, I would appreciate.

See https://github.com/jomjol/AI-on-the-edge-device/pull/996

caco3:
@jomjol Is there something I should change on this PR before you merge it?

caco3:
@jomjol FYI: With the automatic build with the Github Actions, it will not be possible to add the built files to the firmware folder! That's another reason I would get rid of it.

What is the reason you need them in the repo? Maybe we can find a better solution of this. Eg. with the python script I am working on to update automatically.

PR #996, opened 2022-09-10
Auto build and pack

caco3:
Initial Github Action to build firmware and Web Interface.
The version of the Web Interface gets set based on the Git Branch/Commit Hash


caco3:
@jomjol See output in https://github.com/caco3/AI-on-the-edge-device/actions/runs/3028494767
The artifacts always get zipped, it seems not to be possible to disable it. How ever once we change to a single update file, this no longer will be a a problem.

PR #1000, opened 2022-09-10
Fix pipeline

caco3:


caco3:
@jomjol I merge it directly since it is just a fix f the CI/CD pipeline

PR #1011, opened 2022-09-12
Fix precision problem of float variable (6-7 digits)

### LINKED ISSUE #1010: std::stof(NUMBERS[j]->ReturnValue) seems to change the good RawValue, why?
I added some logging..
```
        // Lösche führende Nullen (außer es ist nur noch einen 0)
        while ((NUMBERS[j]->ReturnValue.length() > 1) && (NUMBERS[j]->ReturnValue[0] == '0'))
            NUMBERS[j]->ReturnValue.erase(0, 1);
        LogFile.WriteToFile("delete leading 0: "+NUMBERS[j]->ReturnValue);
        NUMBERS[j]->Value = std::stof(NUMBERS[j]->ReturnValue);
```

2022-09-12T10:29:05: ReturnRawValue: 322924709
2022-09-12T10:29:05: ReturnRaw 32292.4709
2022-09-12T10:29:05: ReturnValue after NaN start32292.4709
2022-09-12T10:29:05: delete leading 0: 32292.4709
2022-09-12T10:29:05: std::stof 32292.47070
2022-09-12T10:29:05: PostProcessing - Raw: 32292.4709 Value: 32292.4707 Value2: 32292 Error: no error

I can see that all is fine until the call of std:stof that normally only converts a string to a float. Maybe a problem of precision?

stefanbode:
fix #1010 
The arbitrary number changes result in an insufficient digit precision of the float definition. With the new double we are at 15-16digits and now a bit better away from rounding effects

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowDefineTypes.h
+++ b/code/components/jomjol_flowcontroll/ClassFlowDefineTypes.h
@@ -33,9 +33,9 @@ struct NumberPost {
     bool checkDigitIncreaseConsistency;
     time_t lastvalue;
     string timeStamp;
-    float FlowRateAct;          // m3 / min
-    float PreValue;             // letzter Wert, der gut ausgelesen wurde
-    float Value;                // letzer ausgelesener Wert, inkl. Korrekturen
+    double FlowRateAct;          // m3 / min
+    double PreValue;             // letzter Wert, der gut ausgelesen wurde
+    double Value;                // letzer ausgelesener Wert, inkl. Korrekturen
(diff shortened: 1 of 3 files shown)


haverland:
A few parameters more should be changed to double.
Look at https://github.com/haverland/AI-on-the-edge-device/commit/1152c917af1ffde3fa5edd40507373ebb5361566 
But beware in the commit the checkDigitIncreaseConsistency is completely removed too.

jomjol:
Included in modified version in the current rolling (20220916)

Issue #1020, opened 2022-09-16
All digits correct - raw value incorrect


tobox:
I am running the latest stable version. All digits are decoded properly:

![Screenshot 2022-09-16 at 10-43-15 watermeter - jomjol - AI on the edge](https://user-images.githubusercontent.com/4719666/190599189-2880dc48-c050-4a32-aa1d-bffb3dc1a1b0.png)

But the raw value is incorrect (5 instead of 6):

![Screenshot 2022-09-16 at 10-42-56 watermeter - jomjol - AI on the edge](https://user-images.githubusercontent.com/4719666/190599332-7c16d89c-af50-4c0c-a11f-9cd70eb453a9.png)

What could be causing this?

tobox:
With the current rolling release it seems to be working correctly!

PR #1021, opened 2022-09-16
Testcases for #950 and #1020

haverland:
- Only test cases from issues. Rolling works correctly here (only float/double value issue)


jomjol:
Manual included in next rolling

PR #1042, opened 2022-09-20
auto release creation and build artifact upload and release notes

haverland:
- automatic release creation by tagging (the master)
- upload old style binary artifacts and new full update.zip
- automatic change of [Unreleased] section to the tag-name (without leading-v)
- automatic release-notes from changelog (the [Unreleased] section)
- after release the updated changelog will be pushed to the master-branch (if tagged a branch, an error occurs). The Changelog must after release manually merged into rolling back.
- changed the release notes to https://keepachangelog.com/en/1.0.0/ style

Best for testing is to create a fork and merge it into master. The 11.3.9 section must be removed after testing. Only for explanation.
Example of created release: https://github.com/haverland/AI-on-the-edge-device/releases


caco3:
@haverland You write that it is used to create a release, how ever it gets called on every push (`on: [push]`). Isnt it supposed to only run on tagging?

P.S. I, at the same time, also improved the script, see https://github.com/jomjol/AI-on-the-edge-device/pull/1043. I will add my changes once yours is merged.

haverland:
The script runs every push to build every time. Look at the if:  only on tagging a release will be build and the changelog changed for the release.

PR #1043, opened 2022-09-20
Update version in autobuild filenames

caco3:
- removed date from filenames
- changed naming format
- split jobs

See https://github.com/caco3/AI-on-the-edge-device/actions/runs/3093802898 for an example output


caco3:
I will wait and rebase on https://github.com/jomjol/AI-on-the-edge-device/pull/1042

PR #1044, opened 2022-09-21
Few typing corrections and translations

cristianmitran:
This PR doesn't contain any code modifications but just a few texts corrections/translations I came across while looking into the code.


jomjol:
Thanks a lot for you effort. 

Only one request: please remove the changes in the directory of the esp32-cam: `/components/esp32-camera-master`
This an external liberay that I use, and the changes will be gone after the next update.

cristianmitran:
I've reversed the changes from the files in that folder.

pos-ei-don:
Why not putting theese Updates in a Patch file and applying this Patch file automated at every build?  Doing so, updates to the  original Files are still possible, in the worst case, this Patch Do not apply cleanly and can be ignored automatically.

caco3:
@pos-ei-don  the better way would be when you could send your typo fixes to the official cam repo.

@jomjol are there several components you copied from other repos? the better way might be to include them as submodules. maybe we can have a look on it in the future when all the ongoing work is completed.

jomjol:
There are 3  module copied:
* esp32-camera-master
* esp-nn
* tflite-lib

caco3:
@jomjol I replaced the folders with submodules, see https://github.com/jomjol/AI-on-the-edge-device/pull/1060

PR #1052, opened 2022-09-21
moved the index-configure.html menus to the index.html page

caco3:
The menu has one level more and we don't need `index-configure.html` anymore:

Screenshot_20220922_000128
---
Screenshot_20220922_000137


caco3:
comming in a separate PR:
- hide menu after click
- improve usage on mobile devices

PR #1060, opened 2022-09-23
Replace components by submodules

caco3:
This PR replaces the following folders with submodules of the corresponding Github repos:
 - `esp32-camera-master` => https://github.com/espressif/esp32-camera.git (pointing to the last release `v2.0.2`)
 - `esp-nn` => https://github.com/espressif/esp-nn.git (pointing to master as there are no tags)
 - `tflite-lib` => https://github.com/espressif/tflite-micro-esp-examples.git (pointing to master as there are no tags)

Since `tflite-lib` is only available as a subfolder of `tflite-micro-esp-examples/`, its location moved to `components/tflite-micro-esp-examples/components/tflite-lib`. I updated `code/CMakeLists.txt` so it can be used as before.

@jomjol The usage is the same as before, you only need additionally to initialize the submodules after a `git clone`, see the new [code/README.md](https://github.com/caco3/AI-on-the-edge-device/blob/7e108d707a03a6a67a448fe301d693928290f27d/code/README.md)!
To update a submodule to a newer version, you can `cd` into that folder and run `git pull`. Make sure to add it to the main repo afterwards!


jomjol:
This is a critical change. There were several releases especially of the esp32cam, where did not work properly or even worse, had an instable system. Only way back was using the old code.
That means, we need to pin all this 3 repositories to a tested version! How can we do that, as there is no versioning there?

caco3:
I saw that you had added zip files with a date in them: 
- `code/components/tflite-lib_20220827.zip`
- `code/components/esp-nn_20220827.zip`
Is this the date you copied the code form the repo? Then we simply can set the submodule commit to that commit.

jomjol:
I tend to decline this change request as I just temporarly bricked 2 out of 4 devices (need to demount and reflash). This is really a problem and we would need to fix it to one very specific version (I even don't know which now).

What is the benefit?

jomjol:
I just tested again and it is working with the current branch. So you could update to this, but still, the risk is much lower, if we don't use the linked, but copied versions.

Your opinion?

caco3:
You have much better control and overview if you use submodules.
For example, I compared your `esp32-camera-master` with many commits from https://github.com/espressif/esp32-camera.git. But I did not find a match! This makes me thinking it could be that you have mixed files from different commits. This would be really bad and call for stability trouble!
With submodules, you easily could switch between commits without manual copying, just point the submodule to another commit.
This makes it much easier to check if a never version is better or worse.

Then I had a look on `esp-nn`. You had a file `code/components/esp-nn_20220827.zip`. The repo got last time updated in May, but that commit again does not match your code base!  Maybe you added that zip file but did not extract it?

jomjol:
Sorry, the date in the file name does not match with any download - it is just for me internal to go back. As just mentioned, I was able to compile it with the newest branches (as of today). 
So if you can lock the submodules to the current version, then it should work and I can test newer version with a redirect.

caco3:
As you can see in https://github.com/jomjol/AI-on-the-edge-device/tree/rolling/code/components, the replaced folders now point to a specific commit of that repo (folder icon with arrow). Unless somebody deletes that repo, there will be no changes on our side when they add a new commit. It will only change through our explicit switch to another commit.

Note that due to the needed rebase one submodule folder got renamed and the other got missing. Not sure how this could happen ,I was just fixing it but you merged already :)
Its fixed now in https://github.com/jomjol/AI-on-the-edge-device/pull/1065

PR #1061, opened 2022-09-24
Update pipeline split jobs

caco3:
- Split the jobs to speed it up
- renamed artifacts

@haverland Can you check I did not break any of your recent changes? In the zipping it complained and suggested to append `-i .`, not sure why. Also it seems you don't upload the generated zip to the artifacts store, is that on purpose?


caco3:
replaced by https://github.com/jomjol/AI-on-the-edge-device/pull/1067

PR #1074, opened 2022-09-25
Release creation like CaCo3's update-ota2

haverland:
- update.zip now like update-rolling
- used initial-setup.zip instead of standalone files
- initial-setup.zip contains now the complete sd-card directory
- updated the Changelog.md in Unreleased part
- suppress extracting of directories in OTA  update

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -181,24 +181,30 @@ jobs:
           ./sd-card/html/version.txt
         key: ${{ github.run_number }}
 
+#########################################################################################
+## Prepare release
+#########################################################################################
+
     - name: Set Variables
       id: vars
       run: |
         echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
 
-    - name: Copy artifacts to firmware folder and create initial_esp32_setup.zip
+    - name: Prepare artifacts  for release
+      if: startsWith(github.ref, 'refs/tags/') 
       run: |
-        mkdir -p firmware
-        # copy builds to firmware folder for committing in next step
-        cp -f "./code/.pio/build/esp32cam/firmware.bin" "firmware/firmware.bin"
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -158,6 +158,18 @@ jobs:
         name: "update__${{ github.ref_name }}_(${{ steps.vars.outputs.sha_short }})"
         path: ./dist/*
 
+    - name: Store generated files in cache
```
caco3, line 161: Unless you modified the files, there is no need to push them to (another) cache!

.github/workflows/build.yaml
```diff
@@ -158,6 +158,18 @@ jobs:
...
+          ./code/.pio/build/esp32cam/partitions.bin
+          ./code/.pio/build/esp32cam/bootloader.bin
+          ./sd-card/html/version.txt
+          ./dist
+        key: ${{ github.run_number }}-pack-for-OTA-v2
+
+
 
 
 #########################################################################################
```
caco3, line 175: `pack-for-fresh-install` doesn't have any use anymore, right?

code/components/jomjol_fileserver_ota/server_file.cpp
```diff
@@ -49,8 +49,8 @@ extern "C" {
 
 /* Max size of an individual file. Make sure this
  * value is same as that set in upload_script.html */
-#define MAX_FILE_SIZE   (2000*1024) // 200 KB
-#define MAX_FILE_SIZE_STR "2000KB"
+#define MAX_FILE_SIZE   (8000*1024) // 8 MB
```
caco3, line 52: I would move this in a separate PR!
maybe combine it with https://github.com/jomjol/AI-on-the-edge-device/pull/1073?

.github/workflows/build.yaml
haverland: I did't saw it. Will change the build structure for cleaner reading.

PR #1086, opened 2022-09-26
Add file length check before upload ota

haverland:
- add check if filename not longer than 30 characters (#1068 )
- fix handling after error dialog 
- added update scenario to release notes
- added "from 11.3.1" files to release artifacts

### DIFF
--- a/sd-card/html/ota_page.html
+++ b/sd-card/html/ota_page.html
@@ -168,6 +168,8 @@ function upload() {
         alert("No file selected!");
     } else if (filePath.length == 0) {
         alert("File path on server is not set!");
+    } else if (filePath.length > 30) {
+        alert("Filename is to long! Use a shorter name.");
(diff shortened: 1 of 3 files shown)


sd-card/html/ota_page.html
```diff
@@ -168,6 +168,8 @@ <h2>Reboot</h2>
         alert("No file selected!");
     } else if (filePath.length == 0) {
         alert("File path on server is not set!");
+    } else if (filePath.length > 30) {
+        alert("Filename is to long! Use a shorter name.");
```
caco3, line 172: @haverland Please provide the number, else people have to do try&error until it fits the check

caco3 (changes requested):

sd-card/html/ota_page.html
haverland: added

PR #1100, opened 2022-09-27
Update Changelog.md

caco3:
@jomjol You should not use `)` for the enumerations, it breaks the formatting as everything then gets put on one line!


haverland:
Was mine. I run to often in this.

PR #1118, opened 2022-09-29
Update Changelog.md

caco3:
I updated the changelog to link to the releases.

Then I realized that the `changelog.md` has that same content as the releases page (https://github.com/jomjol/AI-on-the-edge-device/releases). So in my oppinion we could delete the `changelog.md`, what do you think @jomjol, @haverland ?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,6 @@
 # Changelog
 
-## [Unreleased]
-
-## [12.0.1] - 2022-09-29
+## [12.0.1](https://github.com/jomjol/AI-on-the-edge-device/releases/tag/v12.0.1), 2022-09-29
 
 Improve **u**ser e**x**perience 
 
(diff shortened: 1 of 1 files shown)


jomjol:
> 

I would keep the changelog.md. It is a one page overview over the whole history. There are even some images implemented for an older version.
If we only use the release page, than you need to navigate through different sides if you want to have infos from older version.

caco3:
> I would keep the changelog.md. It is a one page overview over the whole history. 
If we only use the release page, than you need to navigate through different sides if you want to have infos from older version.

Yep, I see that point

> There are even some images implemented for an older version. 

The releases page also supports markdown AFAIK, so the same would be possible. It mainly would reduce the work on a new release.

jomjol:
But anyway. We document the changes in the changelog.md. So it is present and contains all infos, the workflow is working, so why skip it.

haverland:
> I updated the changelog to link to the releases.
> 
> Then I realized that the `changelog.md` has that same content as the releases page (https://github.com/jomjol/AI-on-the-edge-device/releases). So in my oppinion we could delete the `changelog.md`, what do you think @jomjol, @haverland ?

The changelog.md creates the Release notes. During development, the changes are collected here.

PR #1121, opened 2022-10-02
Support meter clock over

cristianmitran:
In case of meter clocking over, that is, reaching its max. value and starting over from 0,
  accept the new value and calculate correctly the difference.
  (see line 739 onwards in ClassFlowPostProcessing.cpp)
  
I don't know enough C to take up the challenge but here is a mathematical approach I would suggest:
1. calculate the number of (rounded) digits of the previous value;
2. calculate the meter's max. value based of the nr. of digits (pow(10d, nrOfDigits));
3. substract previous value from the max value above;
4. add the substraction above to the new value => this will be the difference.

### DIFF
--- a/FeatureRequest.md
+++ b/FeatureRequest.md
@@ -11,6 +11,12 @@
 
 ____
 
+#### #30 Support meter clock over
+
+* In case of meter clocking over, that is, reaching its max. value and starting over from 0,
+  accept the new value and calculate correctly the difference.
+  (see line 739 onwards in ClassFlowPostProcessing.cpp)
+
 #### ~~#29 Add favicon and use the hostname for the website~~- implemented v11.3.1
 
 ~~* https://github.com/jomjol/AI-on-the-edge-device/issues/927~~


jomjol:
I've integrated it into the request list. To be honest, changes on this part of the algo are very critical. There is another approach to go, but this takes quite some time. 
Until this, I don't think, that we will follow two approaches.

cristianmitran:
> I've integrated it into the request list. To be honest, changes on this part of the algo are very critical. There is another approach to go, but this takes quite some time. Until this, I don't think, that we will follow two approaches.

Indeed, I saw it in the code. I started to to give it a go but stopped after a while and reversed. I could do it if someone could help me understand if I can build a kind of a test class to run only on ClassFlowPostProcessing.cpp, without having to "trial and error". My field of expertise is Java and Python, not a C developer, unfortunately.

jomjol:
Send it around and we can have a look together to this.

PR #1159, opened 2022-10-10
Add basic auth filter

crasu:
# Summary

This adds http basic auth in order to have at least a basic level of security.

# Tests

* Compiles
* I can click around in the gui

### DIFF
--- a/code/components/jomjol_controlGPIO/server_GPIO.cpp
+++ b/code/components/jomjol_controlGPIO/server_GPIO.cpp
@@ -27,7 +27,7 @@
 #include "Helper.h"
 #include "interface_mqtt.h"
 
-#include "server_main.h"
+#include "basic_auth.h"
 
 static const char *TAG_SERVERGPIO = "server_GPIO";
 QueueHandle_t gpio_queue_handle = NULL;
--- a/code/components/jomjol_controlGPIO/server_GPIO.cpp
+++ b/code/components/jomjol_controlGPIO/server_GPIO.cpp
@@ -27,6 +27,8 @@
 #include "Helper.h"
 #include "interface_mqtt.h"
 
+#include "server_main.h"
+
 static const char *TAG_SERVERGPIO = "server_GPIO";
 QueueHandle_t gpio_queue_handle = NULL;
 
@@ -439,7 +441,7 @@ void GpioHandler::registerGpioUri()
     httpd_uri_t camuri = { };
     camuri.method    = HTTP_GET;
     camuri.uri       = "/GPIO";
-    camuri.handler   = callHandleHttpRequest;
+    camuri.handler   = APPLY_BASIC_AUTH_FILTER(callHandleHttpRequest);
     camuri.user_ctx  = (void*)this;    
     httpd_register_uri_handler(_httpServer, &camuri);
 }
--- a/code/components/jomjol_controlcamera/server_camera.cpp
+++ b/code/components/jomjol_controlcamera/server_camera.cpp
@@ -8,7 +8,7 @@
 
 #include "ClassLogFile.h"
 
-#include "server_main.h"
+#include "basic_auth.h"
 
 #define SCRATCH_BUFSIZE2  8192 
 char scratch2[SCRATCH_BUFSIZE2];
--- a/code/components/jomjol_controlcamera/server_camera.cpp
+++ b/code/components/jomjol_controlcamera/server_camera.cpp
@@ -8,6 +8,8 @@
 
 #include "ClassLogFile.h"
 
+#include "server_main.h"
+
 #define SCRATCH_BUFSIZE2  8192 
 char scratch2[SCRATCH_BUFSIZE2];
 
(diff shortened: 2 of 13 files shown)


jomjol:
@crasu: first of all thank you very much for your effort and contribution to this project.

But after a short internal discussion I tend not to integrate the basic auth feature into the main branch out of the following reasons:
1. Basic auth is not a very safe way of securing access (I'm not an expert, but rely on other opinions)
Therefore people might think to have a secure remote access and open the esp32 directly into the intranet, which should not be done with a basic auth.
2. If we would like to do it for an external direct access, it would be much more effort and also already some professional part for going into a commercial application. This is clearly not the target for this project.
3. Implementing a security layer also means some extended support and update strategy to scan and close security leaks, because we would always need to close bugs.

crasu:
Thank you for your response. 

If you are concerned that users will start routing the AI-esp over the public internet I can add a parameter to only allow connections from the local subnet.

Btw with the current implementation everyone can read all credentials from the AI-ESP (wlan.ini) via OTA upload of a new firmware and then reading wlan.ini/config.ini (....) file. I am unsure if you are aware of this. 

I just added this because I did not want a completly unsecured device in my local network.

jomjol:
I'm aware of that and therefore you should always run this behind a firewall or an internal network.

pos-ei-don:
But also on m interna network, Devices should hide sensitiv data. Sometimes you have friends in your WLAN or ther Devices. A stranger can just plug in the rj45 in the wall, Google Nest is scanning the whole network. So why not add basic auth and disable it per default? Logins should be logged so external Tools can Block ip's that are trying too often.  As a stranger even can Flash the esp32, He might also install a Firmware which helps him to over take my internal network.

jomjol:
This is a not a professional, but a free firmware. If anybody is afraid, that an attacker could flash the esp32 to get access to the wlan password out of the `config.ini`, than the AI-on-the-Edge tool would be in the wrong environment anyway. 

It is anyhow requested to put sensitive devices in a dedicated and separated network, that is not accessible from outside.

By not providing an authorization etc. algorithm it is at least is clear, what the internal security measures are (none!) and who needs to take care about securing the access (the user!).

I would like to keep it like this.

P.S.: another reason is, that we are adding feature after feature and the effort to maintaining is increasing (see discussion about  InfluxDB, request to test mqtt for the port support). Therefore I would like to limit the development to the technical features in this repo.

crasu:
Will maintain this in my fork. May be you want to remove this from your requested feature list than.

pos-ei-don:
Thank you.

caco3:
> But also on m interna network, Devices should hide sensitiv data. Sometimes you have friends in your WLAN or ther Devices. A stranger can just plug in the rj45 in the wall.

I see your point and agree that the Wifi password is not protected well. How ever I don't see a possible way to protect it securely. We could encryt it (or the partition containing it). But where do we store the encryption key? You can't simply store it in the firmware as it is an opensource project (and on closed source it also would not be very save). The only way i see is if you would have a TPM module, but the ESP32 does not have that!

> To centralize logging (and protection) and Start Action based on errors or Logins, i'd love to be able to log to syslog ;-)

If you can create a pull request for this, we can have a look on it. Since the flash is quite full, we have to be carefully on adding new features.

BTW: 
Basic Auth would also not protect from flashing another firmware as whoever has access to your network simply has to wait until you login with the plain text password!

pos-ei-don:
> > But also on m interna network, Devices should hide sensitiv data. Sometimes you have friends in your WLAN or ther Devices. A stranger can just plug in the rj45 in the wall.
> 
> I see your point and agree that the Wifi password is not protected well. 

I don't understand. When Access is restricted to using Basic-Auth, the Wifi password is restricted aswell.
It is clear, that you can bruteforce this Auth, but normal users won't do that, it needs time and it clearly makes
it illegal. But I agree if you say this is a feature not intended for this project.
I hope that we don't get cluttered by getting too many different forks.

> If you can create a pull request for this, we can have a look on it. Since the flash is quite full, we have to be carefully on adding new features.

i am currently working on a ship, so i dont have permanent access to a syslog server. I will try later this year, when i am back, but cannot say now what my next job will be. I see the flash getting full, but for example i don't need it to frite to influx, so i'd remove this feature for me to get more space free. But I see your point.


> 
> BTW: Basic Auth would also not protect from flashing another firmware as whoever has access to your network simply has to wait until you login with the plain text password!

And how can one easily read my plaintext passwort in a switched network?

caco3:
> And how can one easily read my plaintext passwort in a switched network?

I never did it, but I am confident that it is **very easy** with the right tools and knowledge!

> I don't understand. When Access is restricted to using Basic-Auth, the Wifi password is restricted aswell.

Some background information: https://miloserdov.org/?p=5303

pos-ei-don:
every car can be easily opened. would you buy a car without any lock?
So this argumentation is kind of strange. You can change unsecured things even "accidental" as a guest, but if they are "secured" with basic auth, you at least must have some kind of criminal energy.

caco3:
> every car can be easily opened. would you buy a car without any lock?

I really hope you don't drive a car which got developed by hobbyists in their free time :smile: 

I also would like to have it secured, but fully agree with @jomjol that this is not feasible at the current time with the given resources (available flash, manpower, ...).

PR #1177, opened 2022-10-15
accept minor negative values if extended resolution is enabled

haverland:
- accepts 0.2 differences of negative value only if extended resolution is on (#1176)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,11 +9,12 @@
 ### Changed
 
 -   Updated built environment to `espressif32@v5.2.0`
-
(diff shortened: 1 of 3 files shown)


jomjol:
I don't want negative values, even small ones. Therefore we need to make this configurable.

pos-ei-don:
> I don't want negative values, even small ones. Therefore we need to make this configurable.

+1

haverland:
It will not reduce the value and works like before. Only the error will not set.

PR #1181, opened 2022-10-16
date for graph is now selectable

rdmueller:
another update for the graph feature

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,8 @@
 
 ### Added
 
--   Addes PreValue to `/json` ([#1154](https://github.com/jomjol/AI-on-the-edge-device/issues/1154))
+-   added `/graph.html` to fetch measurements from the debug log and display them as a graph. Activate debug logging for this feature to work.
+-   Added PreValue to `/json` ([#1154](https://github.com/jomjol/AI-on-the-edge-device/issues/1154))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


jomjol:
@rdmueller: there is no a much simpler data format availalbe.
After installing the current rolling, you find in `/log/data` a textfile, that contains only the measurement values (raw, value, prevalue, rate rel, rate abs, read out cnn-numbers). 
This has much more data, than the logfile and is much easier to parse.

jomjol:
@rdmueller: I just adapted the code to data - very easy - GREAT!

View things open:
- in multi number counters (e.g. HT & NT), one would need to select for this as well.
- would be great to know, which data is available, so load all available data files into the preselect.
- there are now also other values available, e.g. rate (abs or relativ) --> could also be selectable

let's see, if I find time to implement a bit of this

rdmueller:
Great to hear that you like this feature.
I hope I can find some time to address those missing features.
Would be great if you could tag your repo or at least my PRs for hacktoberfest 😀

jomjol:
I'm not in the topic hacktoberfest and honestly have no idea, what it is about, and what happens with this.
We have so many mails and request, that I'm not searching for another source of input.

I'm already working on the integration of several numbers and a list of available datas

rdmueller:
@caco3 thanx for adding the label!

PR #1193, opened 2022-10-19
Update build.yaml

caco3:
Increased action versions and migrated to avoid warnings


caco3:
@jomjol @haverland FYI

PR #1203, opened 2022-10-22
Add log level to logfile

caco3:
First merge https://github.com/jomjol/AI-on-the-edge-device/pull/1202!

@jomjol There are some parts I do not understand:
 - [ ] In the config we can set `extended logging` to true/false. The parameter seems to be called [Debug_Logfile_value1](https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/sd-card/html/edit_config_param.html#L1267). How ever I find no reference in the firmware to this! In opposite, there seems to be the parameter [LOGLEVEL](https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_flowcontroll/ClassFlowControll.cpp#L513) which does not get set through the UI!
 - [ ] What about `debugdetailgeneral`? That seems never be set otherwise.

### DIFF
--- a/FeatureRequest.md
+++ b/FeatureRequest.md
@@ -11,6 +11,12 @@
 
 ____
 
+#### #32 Add feature to correct misinterpreted value
+
+* If a value is misinterpreted, the user can manually correct the value.
+* The misinterpreted ROIs would be saved in a "training data" -folder on the SD-card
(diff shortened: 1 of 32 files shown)


caco3:
replaced by https://github.com/jomjol/AI-on-the-edge-device/pull/1206

Issue #1106, opened 2022-09-28
Enhance Log Viewer


caco3:
Enhance the log viewer:
 - [x] It should, by default show the last x lines of the current log
   - https://stackoverflow.com/questions/17877025/reading-last-n-lines-from-file-in-c-c
 - [x] Add button to reload
 - [x] Add links to older and current full file
 - [x] Add colors to errors resp. warnings
 - [x] automatically scroll down
 - [x]  add expert parameter loglevel with a dropdown list DEBUG, INFO, WARNING 
 - [x] update extended logging description

# Nice to have:
 - [ ] automatically reload

caco3:
Idea: Use websocket: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_server.html#websocket-server

caco3:
colorize log:
```html
<p id="log"></p>


<script>

log = "\
01 <INF> info text\n \
02 <WRN> warning text\n \
03 <ERR> error text\n \
04 <DBG> debug text\n"


log = log.replace(/</g, "&lt;").replace(/>/g, "&gt;");
logArr = log.split("\n");
logArr.forEach(processLogLine);

document.getElementById('log').innerHTML = logArr.join("\n");

console.log(document.getElementById('log').innerHTML);



function processLogLine(line, index, arr) {
    if (line.includes("&lt;WRN&gt;")) {
        arr[index] = "<span style=\"color:orange\">" + line + "</span>";
    }
    else if (line.includes("&lt;ERR&gt;")) {
        arr[index] = "<span style=\"color:red\">" + line + "</span>";
    }
    else if (line.includes("&lt;DBG&gt;")) {
        arr[index] = "<span style=\"color:gray\">" + line + "</span>";
    }
//     else if (line.includes("&lt;INF&gt;")) {
//         arr[index] = "<span style=\"color:blue\">" + line + "</span>";
//     }
    
    arr[index] += "<br>";
}
</script>
```

caco3:
A new URL got added: `http://<IP>/log`. It loads only the latest 80 kBytes of the current log

caco3:
merged with https://github.com/jomjol/AI-on-the-edge-device/pull/1243

PR #1206, opened 2022-10-22
Add log level to logfile

caco3:
@jomjol There are some parts I do not understand:
 - [x] In the config we can set `extended logging` to true/false. The parameter seems to be called [Debug_Logfile_value1](https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/sd-card/html/edit_config_param.html#L1267). How ever I find no reference in the firmware to this! In opposite, there seems to be the parameter [LOGLEVEL](https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_flowcontroll/ClassFlowControll.cpp#L513) which does not get set through the UI!
 - [x] What about `debugdetailgeneral`? That seems never be set otherwise.
 
# TODO
 - [x] Test on device

### DIFF
--- a/code/components/jomjol_controlGPIO/server_GPIO.cpp
+++ b/code/components/jomjol_controlGPIO/server_GPIO.cpp
@@ -400,7 +400,7 @@ bool GpioHandler::readConfig()
 
     if (gpioExtLED > 0)
     {
-    //     LogFile.WriteToFile("Startsequence 06");      // Nremove
+    //     LogFile.WriteToFile(ESP_LOG_INFO, "Startsequence 06");      // Nremove
 //        vTaskDelay( xDelay );   
 //        xDelay = 5000 / portTICK_PERIOD_MS;
 //        ESP_LOGD(TAG_SERVERGPIO, "main: sleep for: %ldms", (long) xDelay);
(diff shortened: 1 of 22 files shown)


jomjol:
> In the config we can set extended logging to true/false. The ...

This UI setting changes the setting of  `void SwitchOnOff(bool _doLogFile)` in the `ClassLogFile.h`
This only switches the external log file on/off. We can extend this somehow to the Log-Level or we make another variable in the expert session to controll the serial log level - I would prefer the second option.



>  What about debugdetailgeneral? That seems never be set otherwise.

This is set inside the file `ClassFlowCNNGeneral` in the protected part of the class `ClassFlowCNNGeneral`:
```
protected:
    bool debugdetailgeneral = true;
    t_CNNType CNNType;
```

caco3:
Thanks!
I also was thinking of removing the `extended logging` from the UI and instead add a dropdown list where one can select DEBUG, INFO, WARNING. And use INFO as default. 
I would only use it for the log file and keep the UART log always on DEBUG.

I can see that `debugdetailgeneral` gets set to true once, so IMO could be replaced by above level switch

caco3:
@jomjol Please merge it as is now. I will add the debug level selection in a later PR related to https://github.com/jomjol/AI-on-the-edge-device/issues/1106

PR #1216, opened 2022-10-24
Update build.yaml

caco3:


caco3:
@jomjol This should now also allow building on PR from forks

PR #1219, opened 2022-10-25
Fix Github-action on pull requests

haverland:
- pull requests containing slashes that will break the filename convention 
- replaces slashes by dot


caco3:
thanks, it was too late yesterday to fix it myself 😒

PR #1231, opened 2022-10-27
Improve logging to logfile

caco3:
An error log message now always gets logged to the logfile, even if disabled.
This also means we no longer need to enable the logfile to log an error.


caco3:
@haverland FYI

PR #1232, opened 2022-10-27
Improve MQTT

caco3:
- Please first merge https://github.com/jomjol/AI-on-the-edge-device/pull/1231
- Please squash the PR (always suggested to do on my PRs :smile:  )

### Changes
- Refactored MQTT implementation
- Use hostname as Client ID
- Added topics containing IP, hostname and MAC

### Open questions
- [x] @jomjol I need to know the `AutoIntervall` in https://github.com/jomjol/AI-on-the-edge-device/pull/1232/files#diff-8b3f08497113b4639a307ee14e7bcfa356ac3f750c8cb8659da7a7b244f33d62R93, but I could figure out out to access `ClassFlowControll` as it is not in the `lfc` list.

Sporadically, the initialization still fails, but on the next round it successes, so that is not a huge problem.
Whenever it connects, it sends the static information like IP, MAC, ...


jomjol:
"Autointervall" is the internal storage of the "Intervall" parameter = cycle time of the recognition flow. If it would ever not be set by the config.ini it`s default value would be "10" = 10 minutes.

caco3:
@jomjol Yes I know. I need it for the MQTT Timeout. I want to set it to 2.5 * this interval. This way we don't get immediately a "Device disconnected" when one round fails to send MQTT.

So is there a way to access it through the `ClassFlowControll` or should I let it read the config itself? i would prefer the first one so it uses the same default value defined on one place.

caco3:
merged to avoid conflicts

Issue #1180, opened 2022-10-16
What about a Webinstaller?


helllth:
### Wouldn't it be cool to lower the hassle of the initial installation with a webinstaller?  

<img width="1044" alt="Screenshot 2022-10-16 at 12 10 35" src="https://user-images.githubusercontent.com/837211/196029675-b055c874-6a80-40a6-85a1-f6fdcf25f887.png">

I forked the repo to https://github.com/helllth/AI-on-the-edge-device and added a `gh_pages` branch to publish a slightly changed version of the `Readme.md `on https://helllth.github.io/AI-on-the-edge-device/.

Couldn't make a pull request though because there is no `gh_pages` branch in the current repo. (Therefore this issue).

Hope you'll like the idea :wink:

### Version

any

### Expected Behavior

You will find the Install Button on https://helllth.github.io/AI-on-the-edge-device/index.html#install below the **Download** section.

You'll just have to connect your device with an USB cable to your computer and click the button:

![Screen Recording 2022-10-16 at 11 51 49](https://user-images.githubusercontent.com/837211/196029819-19a91530-e434-469c-a6e0-6be890d140e8.gif)

Furthermore you can connect to an existing AI device and watch what is going on on the console (This example is from a hardware reset.):

![Screen Recording 2022-10-16 at 11 55 21](https://user-images.githubusercontent.com/837211/196029972-7de95c70-cc6d-4c78-a751-7353058b1471.gif)

### Additional Context

 **If you like the Idea of a Webinstaller, please add a `gh_pages` branch to your project and remove all content except the images directory. After that I can make a pull request to update that branch.** 

The pull request will contain an index.html that I manually converted from the current `Readme`.md with an added section **Webinstaller and Console** with the magic button.

(Before you merge the pull request you have to activate the `gh_pages` though - Deploy from a branch).

<img width="1243" alt="Screenshot 2022-10-16 at 12 07 18" src="https://user-images.githubusercontent.com/837211/196029559-8d8b10a2-b1f4-443b-8512-ef92bbfbddfb.png">

jomjol:
In principle like the idea. Two points:
1. manual sd-card writing is still unavoidable
2. I develop everything with firefox, this is not supported to my knowledge
3. is there a change to choose between two configurations? e.g. rolling and master?

caco3:
Some comments to above points:
[...]

PR #1245, opened 2022-10-30
Webinstaller and Console for Browsers that support Web Serial API (Chrome, Edge and Opera)

helllth:
### Wouldn't it be cool to lower the hassle of the initial installation with a webinstaller?  

Screenshot 2022-10-16 at 12 10 35

You'll just have to connect your device with an USB cable to your computer and click the button:

Screen Recording 2022-10-16 at 11 51 49

Furthermore you can connect to an existing AI device and watch what is going on on the console (This example is from a hardware reset.):

Screen Recording 2022-10-16 at 11 55 21

(Before you merge the pull request you have to activate the `gh_pages` though - Deploy from a branch).

Screenshot 2022-10-16 at 12 07 18

see further discussions in https://github.com/jomjol/AI-on-the-edge-device/issues/1180

--- 

This is just a very first try to add the cool https://esphome.github.io/esp-web-tools/ Button. It does not need to be a copy of the `Readme.md`. Could be also just a **Console and Install** site. 

To experiment with direct references to the releases I would recommend an `experimental` or `alpha` folder to avoid breaking things.

Hope you'll like it. :metal:


manifest.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "AI-on-the-edge",
+  "version": "12.0.1",
+  "funding_url": "https://www.paypal.com/donate?hosted_button_id=8TRSVYNYKDSWL",
+  "new_install_prompt_erase": false,
+  "builds": [
+    {
+      "chipFamily": "ESP32",
+      "parts": [
+        {
+          "path": "releases/download/v11.2.0/bootloader.bin",
```
caco3, line 11: Instead of having to add the binaries to the repo, can we refer to an URL, eg. https://github.com/jomjol/AI-on-the-edge-device/releases/download/v12.0.1/firmware.bin?
If this would work it would make maintenance easier. The needed binaries can be added to the release if needed.

helllth: I had a hard time with referencing them directly from my fork. (Cross origin stuff)

You could an alpha folder were you could experiment with the direct links in the manifest though.

jomjol:
Is there a reason, why you commit to jomjol:gh_pages?

Could I push to jomjol:rolling directly?

jomjol:
Maybe I did not understand the concept behind the web-installer. Therefore following questions

- [ ]  Do we need a gh_pages branch or can we add this to the rolling and later to the main branch?
- [ ]  I would like to reduce the doubled information as much as possible. Is it possible to make a html-page with only the installer and keep all other informations in the `readme.md`?

caco3:
> You need the gh_pages branch

I think (and would prefer), if we would use a subfolder of the branch `rolling` resp. `master`:
grafik

jomjol:
I agree with @caco3 and I thought I tried to use a subfolder in the main / rolling path. I will rename the gh_pages for a test in the next days.

Issue #1133, opened 2022-10-05
MQTT Discovery in Homeassistant


caco3:
- [x] Implement
 - [x] Update wiki page https://github.com/jomjol/AI-on-the-edge-device/wiki/Integration-Home-Assistant

caco3:
See https://github.com/jomjol/AI-on-the-edge-device/issues/1108

dcboy87:
okay, i also played a bit today, maybe the problem is formatting of the JSON payload. I tried with this example:

````
{
  "~": "watermeter",
  "name": "Meter-Water - Counter",
  "unit_of_meas": "m³",
  "icon": "mdi:water",
  "state_topic": "~/main/json",
  "value_template": "{{ value_json.value}}",
  "json_attributes_topic": "~/main/json",
  "unique_id": "smartmeter.watertotal.total",
  "availability_topic": "~/connection",
  "payload_available": "connected",
  "payload_not_available": "connection lost",
  "device": {
    "identifiers": ["12345678"],
    "name": "Total water meter",
    "model": "M100",
    "manufacturer": "aquamess"
    }
  }
````

then in HA this device gets discovered:

![grafik](https://user-images.githubusercontent.com/45015463/194227591-331de1b2-4dc2-435a-8d64-d3277e5e44b3.png)
![grafik](https://user-images.githubusercontent.com/45015463/194227834-1a5663eb-fc1c-4861-b159-98c678894261.png)

**Comment by caco3** [comment]
thanks for testing it!
I tried to use your example in the HA MQTT sending pat, but that did not work at all (no message gets sent at all, monitored with http://mqtt-explorer.com).
Then i sent the very same message/topic using http://mqtt-explorer.com. Then suddenly HA saw it and showed the new device.

So it seems to work in general.
I will have a look on in some weeks (am on holiday right now).

**Comment by caco3** [comment]
I now got it working with a Python MQTT client.
I will try to implement it on ESP side after my holidays.

**Comment by dcboy87** [comment]
thank you very much! Great work 💪

**Comment by Liese2007** [comment]
Hi,
have actually problems getting my sensor to transmit data to Home Assistant via mqtt broker.
Is my understanding right, that I have actually to wait for next release?

Here is my log, just in case my fault is a different one:
```
DigitalUnschaerfe = 0.200000
2022-10-19T20:38:45: ClassFlowCNNGeneral::getReadout#ZeigerEvalHybridNeu()= 1
[...]

PR #1246, opened 2022-10-30
### DELETE ### Enhance Homeassistant Discovery

caco3:
- Added parameter to enable/disable discovery
- Added parameter to select meter type. This is used to set the `device_class` and `unit_of_meas` correctly
- Corrected various Discovery parameters, use `diagnostic` in the `entity_category` for the diagnostic sensors

### New Topics
 - `interval`: Interval of the rounds in minutes as configured in the config.

### Notes
 - The Discovery Topics and the static topics (`IP`, `MAC`, `Hostname`, ...) only get sent on startup. To trigger a re-sending of them, you can call the following URL: `http://<IP>/mqtt_publish_discovery`


caco3:
Waiting for feedback from https://github.com/jomjol/AI-on-the-edge-device/issues/1133

PR #1255, opened 2022-11-01
Cleanup REST API

caco3:
Added new, consistent REST APIs:
- `value`, same as `value.html`
- `setPreValue`, same as `setPreValue.html`
- `statusflow`, same as `statusflow.html`
- `cpu_temperature`, same as `cputemp.html`
- `rssi`, same as `rssi.html`
- `statusflow`, same as `statusflow.html`

### DIFF
--- a/code/components/jomjol_tfliteclass/server_tflite.cpp
+++ b/code/components/jomjol_tfliteclass/server_tflite.cpp
@@ -837,7 +837,7 @@ void register_server_tflite_uri(httpd_handle_t server)
     camuri.user_ctx  = (void*) "Light Off";
     httpd_register_uri_handler(server, &camuri);
 
-    camuri.uri       = "/cputemp";
+    camuri.uri       = "/cpu_temperature";
     camuri.handler   = handler_cputemp;
     camuri.user_ctx  = (void*) "Light Off"; 
     httpd_register_uri_handler(server, &camuri);  
--- a/code/components/jomjol_tfliteclass/server_tflite.cpp
+++ b/code/components/jomjol_tfliteclass/server_tflite.cpp
@@ -818,7 +818,7 @@ void register_server_tflite_uri(httpd_handle_t server)
 
     camuri.uri       = "/doflow";
     camuri.handler   = handler_doflow;
-    camuri.user_ctx  = (void*) "Light Off"; 
+    camuri.user_ctx  = (void*) "Light Off";  
     httpd_register_uri_handler(server, &camuri);
 
     camuri.uri       = "/statusflow.html";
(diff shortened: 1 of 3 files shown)


caco3:
> Okay for me but we need to state clearly, that it is not downward compatible then.

Yes, for sure!
Also MQTT will have some breaking changes.

Therefore it might be wise to again increase the version to `14.0.0`.

haverland:
I think it's not a good idea to break APIs, especially MQTT. That breaks for the most users and don't want reconfigure all devices.

jomjol:
What does change in the MQTT? can it be configured in that way, that it only changes for the homematic users?

caco3:
I would like to change the REST API, because it is not consistent and confusing.

As for MQTT, there is the same reason. See https://github.com/jomjol/AI-on-the-edge-device/pull/1246 for details. As for MQTT it helps to get a better understanding of what the topic contains. And most changes are not used for automations or long term data recording, so that should be fine.

jomjol:
To be honest, I also have problems with the change of the MQTT interface. It is not only, that I need to adjust the transfer to the database. But I'm using iobroker as my system and a lot of controll loops base on the input from the mqtt.

E.g. is my house temperature controll system is interlinked with the power meter of the photovoltaik system. There is a script, that compares the power rate with the expected power rate, out of this I judge with an algorithm, if I want my sun blinds to go up or down. All this is hidden in quite some old javascripts, I don't want to change. 

And this is only one example. Therefore I also need a time of downwards compatiblity in order to slowly adjust and test my scripts.

caco3:
So should I add for both add a `legacy_API` parameter?

jomjol:
Yes, that would give everybody the change to change until the overnext release his internal procedures.

jomjol:
For new installations, this is okay for sure. If I do an upgrade I would prefer to stay automatically in the current settings.
We could do this by adding a paramter into the `config.ini` that is set to false (e.g. `legacy_mode`). If not present, the firmware sets it to true.

caco3:
I see your point, but I fear more questions from existing users because they do not understand the old naming.
Why not make the new one default (new and existing setups).
When we note the break in the readme and point to the parameter, it should be easy for those who **can** not switch.

I don't think that we ever can get rid of the old (current) API, or would you ever want to update your scripts?

As for the parameter: 
I would not use a boolean value, just in case we will need a future API break!
I would use a version field in the `[System]` section:
`api_version = latest`
- By default it is set to `latest` on new config files.
- And for legacy support, one can set it to 12 (last version where it was the default).

If you can't agree with the default on existing systems, the software would need to set it to `12` if it does not find it.

caco3:
@jomjol I decided to not replace the REST API but instead add new, redundant APIs.

For MQTT I will add a new API which can be configured.

Please have a look and merge

PR #1268, opened 2022-11-04
Fix various warnings which become fatal in later versions

bzfbd:
- we cannot use partial initialisation of structs in C++ files (copied from example C files initially it seems)
- IRAM_ATTR uses a COUNTER, do not use the attribute on the implementation
- provide missing copy implementations for Rgb and Hsv
- one no longer can |= on volatile variables; use = | instead
- fix project and header includes
- avoid redefining BLINK_GPIO
- Remove defined but unused variables
- Fix printf formats (not sure why they change in 5.x)
- Add missing case statement for 5.x (HTTP_EVENT_REDIRECT)
- RMT needs to be updated to new interface (CONFIG_RMT_SUPPRESS_DEPRECATE_WARN is on currently; see https://docs.espressif.com/projects/esp-idf/en/release-v5.0/esp32/api-reference/peripherals/rmt.html)
- Adjust tcpip_adpater_* to esp_netif_*
- Use buffered versions of *ntoa* functions for IPv4 addresses and not a static on the stack (also fixes warnings)
- Whatever I missed


jomjol:
sorry, need to revert the commit. 
The system is not stable anymore (reboot during MQTT). Can you commit the changes in smaller parts? Changing 25 files is hard to debug, if it is not working.

PR #1270, opened 2022-11-04
Correct spelling of "Hostname"

bzfbd:


caco3:
hmm, odd, this type does not exist anymore, so no need to create a new PR!

PR #1312, opened 2022-11-09
Update web installer

caco3:
- Updated the web Installer page
- removed all redundant files in the docs folder
- updated the main README


jomjol:
you should not remove `sd-card/html/graph.html`. It is neede for the graphical draw function.

caco3:
> you should not remove `sd-card/html/graph.html`. It is neede for the graphical draw function.

The one in the docs folder? Are you sure?

jomjol:
You're right - did not realise the /doc, just saw /sd-card/html ;-)

PR #1339, opened 2022-11-12
Lock sizes and keep X spaces equidistant

caco3:
@jomjol Please test and let me know if I misunderstood something.

Hopefully this prevents some users to have misaligned setups

### General
- Changed `d` to `Δ`. For example `dx` => `Δx`
### Digital
- Add button to make the X space between all values equidistant
- Add checkbox (default enabled) to lock `y`, `Δx` and `Δy` to be identical between all ROIs
- Show warning when lock gets unchecked.

### Analog
- Add checkbox (default enabled) to lock `Δx` and `Δy` to be identical between all ROIs

### DIFF
--- a/sd-card/html/edit_analog.html
+++ b/sd-card/html/edit_analog.html
@@ -107,7 +107,7 @@ th, td {
 		</td>
 		<td>                
             <input class="button" type="submit" id="renameROI" name="renameROI" onclick="renameROI()" value="Rename">  
-		<td>
+        </td>
 		<td>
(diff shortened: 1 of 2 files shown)


jomjol:
I tested and like it very much. Only the Make x-distance equivalent might be a bit non intuitive to push this button
Greatm work, thank!

caco3:
Thanks!

About the X equidistant (not equvalent :smile: ), its a bit trick to have it auto, because if you move it around manually, to which other number should it be referenced?
Also, when you add a new number, it has x/y coordinats = 1. This could be solved, I will have to have a look.

PR #1396, opened 2022-11-22
slow down reboot loops

caco3:
- If a device restarts due to an `exception/panic`, the first round gets delayed by 5 minutes. this gives time to fetch the log or run an OTA DFU. Also, the log level gets set to DEBUG until the next start. 
- If cam initialization fails, it retries once. Usually it successes then.
- If the initialization of the cam, NTP or cam framebuffer fails, it waits 5 minutes and restarts then.


caco3:
Since a software restart also causes a `Exception/panic` due to `Failed to initialize camera module. Check that your camera module is working and connected properly.`, we only should delay when we have 2 `Exception/panic`s in a row => store counter on SD card and reset on any other restart reason

PR #1440, opened 2022-11-30
Improve mqtt

caco3:


caco3:
@Slider0007 Thanks for your suggestion! Could you create a Pull request for it?

Some points:
- The stability issues we have are seldom really on the initial init AFAIK see, but your branch is an improvement anyway.
 - Yes, the heap logging should be disabled in general, I had it enabled to try to find out why MQTT sometimes fails, but it does not seem to be a memory problem.
 - The retrials which you commented out: I saw a few times that it failed one time and succeeded on a retry, not sure why, they are Espressif functions and there is no further error explanation :unamused: So I would keep them there.
 - I still do not understand why we have random disconnects or no disconnect but failing publishs :unamused:

Issue #1455, opened 2022-12-01
MQTT improvements


caco3:
From @slider0007
 @caco3: It seems that during initial startup of the device the MQTT init service is called multiple times in a short time due to the fact that the call is in the ReadParameter routine. Moving the MQTT init service after the initializiation of the flow classes are done at least the errors after initial startup are gone.

Please have a look the this branch maybe this helps somehow --> [Branch](https://github.com/Slider0007/AI-on-the-edge-device/tree/mqtt-errors)

_Originally posted by @Slider0007 in https://github.com/jomjol/AI-on-the-edge-device/issues/1440#issuecomment-1333975098_

caco3:
@Slider0007 Thanks for your suggestion! Could you create a Pull request for it?

Some points:
- The stability issues we have are seldom really on the initial init AFAIK see, but your branch is an improvement anyway.
 - Yes, the heap logging should be disabled in general, I had it enabled to try to find out why MQTT sometimes fails, but it does not seem to be a memory problem.
 - The retrials which you commented out: I saw a few times that it failed one time and succeeded on a retry, not sure why, they are Espressif functions and there is no further error explanation :unamused: So I would keep them there.
 - I still do not understand why we have random disconnects or no disconnect but failing publishs :unamused:

Slider0007:
Thanks for your feedback. I create a draft pull request for this proposal.

> * The stability issues we have are seldom really on the initial init AFAIK see, but your branch is an improvement anyway.
> * The retrials which you commented out: I saw a few times that it failed one time and succeeded on a retry, not sure why, they are Espressif functions and there is no further error explanation 😒 So I would keep them there.
> * I still do not understand why we have random disconnects or no disconnect but failing publishs 😒

My thoughts:
- In the recent versions (I cannot tell when it started) I realized that at startup the mqtt errors occur (almost) at every start. The errors at start are gone, when I reduced the calls of the mqtt init routine by removing the retries and moving the inital call after the init of flow classes is done. I assume the call frequency of init is too high and system is not able to react (also no time "event connected" can be triggered). 
- Whenever the WIFI connection is on the edge (very bad connection) I faced mqtt disconnects / publishing errors during runtime as well. Could it be that a not stable / or not good engough WIFI connection is the root cause of random disconnects / publishing issues at runtime? 

[...]

PR #1487, opened 2022-12-04
MQTT improvements

caco3:


caco3:
@jomjol Is it ok to increase the reserved internal memory? It looks like as if the MQTT init fails because FreeRTOS fails to create a new tread for it. How ever I am not sure on the effect on the general memory usage.

See:
 - https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-spiram-malloc-reserve-internal
 - https://github.com/jomjol/AI-on-the-edge-device/issues/1455#issuecomment-1336450501
 - https://esp32.com/viewtopic.php?t=4985

Issue #1468, opened 2022-12-03
json/CSS dispalying problem


jurasjo:
### The Problem

Hi
Thanks for this great project. I use it wit HA and the newest version  works very well. I wanted to let you know there might be a small problem with displaying json value inside "Diagnostic" card in HA. It gets outside of the card (maybe json string should have spaces). 
So now is:
`{"value":58.616,"raw":"058.616","error":"no error","rate":0.000000,"timestamp":"2022-12-03T09:44:24+0200"}`
and probably should be:
`{"value":58.616, "raw":"058.616", "error":"no error", "rate":0.000000, "timestamp":"2022-12-03T09:44:24+0200"}`

Looks like now, space between words "no" and "problem" is the only thing which allows this string to go to the new line.

### Version

Release: v13.0.3 (Commit: e9e1358+)

### Logfile

```shell
na
```

### Expected Behavior

json string should be within the card edges

### Screenshots

![Zrzut ekranu z 2022-12-03 09-37-11](https://user-images.githubusercontent.com/19259405/205430887-93814631-ac96-4c41-856a-dbf36af2cc5d.png)

caco3:
The JSON string is not intended to be used directly in the UI. And the JSON specification does not require or enforce whitespaces outside the keys/values!

Therefore I do not see this as a bug.

jurasjo:
**So it is a HA/Mosquito broker bug??** 

I use this new feature: "Using MQTT (Automatically Setup Entities using Homeassistant MQTT Discovery)". 
Settings > Integrations > MQTT(Mosquitto broker) there is my watermeter and those two cards "Sensors" and "Diagnostic".

caco3:
No, it is not a bug at all!

The discovery page simply shows all discovered entities of our device, so you know what is available. A JSON entity newer is intended to be show directly in the UI, instead it is used for automation's or template sensors.
All information is available separate in other entities which can be added to the UI directly.

jurasjo:
But it is displayed by default. That is the point.

jurasjo:
So here is your wiki page: https://github.com/jomjol/AI-on-the-edge-device/wiki/Integration-Home-Assistant

You are telling me that if we had the whole screenshot there, there would be the same problem as on my screenshot? And it would be normal?

caco3:
you are aware that the screenshot there is cut off? It is only there for illustration!

caco3:
[...]

PR #1488, opened 2022-12-04
add whitespaces in JSON in MQTT

caco3:
See https://github.com/jomjol/AI-on-the-edge-device/issues/1468

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
@@ -225,6 +225,7 @@ bool ClassFlowMQTT::doFlow(string zwtime)
     std::string result;
     std::string resulterror = "";
     std::string resultraw = "";
+    std::string resultpre = "";
     std::string resultrate = ""; // Always Unit / Minute
     std::string resultRatePerTimeUnit = ""; // According to selection
     std::string resulttimestamp = "";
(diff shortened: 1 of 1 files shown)


gorddel:
@caco3 
The whitespace added to JSON string published to MQTT by this change are extremely problematic for me and have caused me to revert to Version 12.0.1. When Mosquitto clients subscribe to that topic, they receive the JSON string broken into newlines at each space. This makes it impossible to handle that JSON string in other scripts. Please revert this change to produce industry standard JSON MQTT strings.
(Sorry, I'm a github novice and I'm not sure if replying to a closed issue is the proper way to address this, or if I should open a new issue).

caco3:
Hi @gorddel 

Yes, it is the right place to refer to  a specific change.

Can you please refer me to that mentioned industry standard?
To my knowledge, whitespaces are allowed. See also https://stackoverflow.com/questions/4150621/are-whitespace-characters-insignificant-in-json where it even refers to an RFC! Also, testing it with shows it is valid: https://jsonformatter.curiousconcept.com/?data={%20%20%20%22key%22:%20%20%20%20%20%20%20%20%22value%22%20%20%20%20%20}&process=true# (on the right side you can even select the RFC to check for, they all show it as valid).

I also use Mosquitto and do not see any issue with it.
So, no, I do not see why we need to change it on our side, please have a look on your script side! Maybe the string is just wrapped around because it is too long?

gorddel:
Hi @caco3 

Thanks very much for your quick reply. Perhaps my phrase of "industry standard" was overstated - you are correct that spaces are allowed in the JSON strings. However, of the many MQTT devices I that use, Tasmota, Zigbee2MQTT, etc, most of them format JSON strings for MQTT without additional spaces. 

I'll dig deeper into things on my side and see how I can handle this change on my end going forward. 

Thanks again for your input and consideration. Jomjol, you, and the others that have created, enhanced, and support this amazing project deserve high praise indeed! I'm awestruck by the amount of functionality that you have packed into an ESP32!

caco3:
Thanks for your undertsanding!

They most likely remove the whitespace to save a bit of memory.
I on purpose added the whitespace so a client is able to wrap the string when it is too wide.
Example:
grafik

Without the whitespaces, you only would see the first part of the whole message-

Issue #1390, opened 2022-11-21
wlan.ini not properly protected


caco3:
from https://github.com/jomjol/AI-on-the-edge-device/discussions/1389

Hi guys,

So I had the issue that I needed to change the SSID remotely with no access to the SD-card but as mentioned several times in the forum this seems to be not possible.

Inventive how I am I made it happen. Not sure if this is considered a bug but at least you cannot read the existing file with the password.
I did this now 2 times already so it definitely works.

You cannot see or read the wlan.ini through the fileserver UI, also you cannot overwrite it with the fileserver UI, it will tell you a file with the same name already exist when you try to upload such a file.

So the trick is that you first need to delete the existing wlan.ini file and then upload a new file through the fileserver UI with the new WiFi settings.

1. Prepare your new wlan.ini file, open Filseserver UI in root and select the File to upload (do not press upload yet)
2. Run the following POST request, e.g. with CURL, powershell, etc. from your PC:
`curl "http://<watermeterIP>/delete//wlan.ini" -X "POST" -H "Content-Length: 0" -H "Content-Type: application/x-www-form-urlencoded"`
You should get the return "File successfully deleted"
3. Upload your new wlan.ini file through Fileserver UI that you prepared, it will succeed this time
4. Reboot through UI and pray that you entered the right SSID and credentials in the file

_Originally posted by @Terr4 in https://github.com/jomjol/AI-on-the-edge-device/discussions/1389_

caco3:
https://github.com/jomjol/AI-on-the-edge-device/pull/1492#pullrequestreview-1205291896

PR #1509, opened 2022-12-07
Fix for securing wlan.ini

parhedberg:
This SHOULD prevent from deleting the wlan.ini file with URL request as of #1390 

Have not been able to test it properly..

### DIFF
--- a/code/components/jomjol_fileserver_ota/server_file.cpp
+++ b/code/components/jomjol_fileserver_ota/server_file.cpp
@@ -829,7 +829,7 @@ static esp_err_t delete_post_handler(httpd_req_t *req)
             return ESP_FAIL;
         }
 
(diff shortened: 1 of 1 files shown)


code/components/jomjol_fileserver_ota/server_file.cpp
```diff
@@ -829,6 +829,12 @@ static esp_err_t delete_post_handler(httpd_req_t *req)
             return ESP_FAIL;
         }
 
+        if (filename == "wlan.ini") {
+            ESP_LOGE(TAG, "Trying to delete protected file : %s", filename);
+            httpd_resp_send_err(req, HTTPD_500_INTERNAL_SERVER_ERROR, "Not allowed to delete wlan.ini");
```
caco3, line 834: You should use code 403

PR #1546, opened 2022-12-11
ENABLE_MQTT c++ macro definition

nliaudat:
Added C++ macro (#ifdef ENABLE_MQTT) to be able to completely disable MQTT and later other features (Influx, GPIO, etc). 

This allows a more thorough debugging and especially to free space for future implementations (like Matter or Ipv6)

To enable or disable MQTT, you have to play with platformio.ini : 
build_flags = -D ENABLE_MQTT

### DIFF
--- a/code/components/jomjol_controlGPIO/CMakeLists.txt
+++ b/code/components/jomjol_controlGPIO/CMakeLists.txt
@@ -4,6 +4,6 @@ list(APPEND EXTRA_COMPONENT_DIRS $ENV{IDF_PATH}/examples/common_components/proto
 
 idf_component_register(SRCS ${app_sources}
                     INCLUDE_DIRS "." "../../include"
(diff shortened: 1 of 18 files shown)


nliaudat:
Original :

RAM:   [==        ]  20.5% (used 67328 bytes from 327680 bytes)
Flash: [==========]  97.7% (used 1901437 bytes from 1945600 bytes)

No MQTT : 

RAM:   [==        ]  20.4% (used 66964 bytes from 327680 bytes)
Flash: [==========]  95.3% (used 1853973 bytes from 1945600 bytes)

jomjol:
Okay - but regarding new features we are very strict. The code is already very complex (for such a low power device) w/o good debugging options.

nliaudat:
Thanks ! 
Macro definition is only usefull for developpers. Debugging and testing new features get simplified
It could be great if every new feature adopt that style of coding. 
Regards

caco3:
It was also on my list to be able to disable certain features this way.

Issue #1551, opened 2022-12-11
Improve handling of insufficient PSRAM


caco3:
### Issue

Until 13.0.5, the device did not continue starting if the PSRAM did not have 4 MBytes or more.

The 4 MBytes are needed, else it will not work!. How ever we want to try to improve the handling of it.

I created a new firmware which tries to give a better output. This will not solve your issues with the device, but helps us to trace it down.

https://github.com/jomjol/AI-on-the-edge-device/actions/workflows/build.yaml?query=branch%3Acontinue-boot-on-bad-psram

Can you please try to run above firmware and show us the (complete) USB log?

caco3:
@Maddin4000

caco3:
@mundi86

Maddin4000:
@caco3 wie installiere ich es?

caco3:
Du musst bei obigem Link auf den obersten Eintrag mit dem grünen Kreis klicken, dann nach unten scrollen und `
initial_esp32_setup__continue-boot-on-bad-psram_*` herunterladen.

Danach installieren wie normal, siehe https://github.com/jomjol/AI-on-the-edge-device/wiki/Installation#flashing , Punkt 2 oder 3.

mundi86:
hi, thx 4 the support.
i flashed the new firmeware
![grafik](https://user-images.githubusercontent.com/13553662/206922916-4e72ff50-1231-4a00-a620-c7dcf5d63448.png)

now the Seriel Monitor shows me more, but it its in boot loop.
here the output:
-------------------------------------------------------------------
ets Jul 29 2019 12:21:46

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:6656
load:0x40078000,len:15300
ho 0 tail 12 room 4
load:0x40080400,len:3848
entry 0x40080694
[0;32mI (29) boot: ESP-IDF 4.4.2 2nd stage bootloader[0m
[0;32mI (29) boot: compile time 17:05:52[0m
[0;32mI (29) boot: chip revision: 3[0m
[0;32mI (32) boot_comm: chip revision: 3, min. bootloader chip revision: 0[0m
[...]

PR #1583, opened 2022-12-13
extended instructions

caco3:


caco3:
See https://github.com/jomjol/AI-on-the-edge-device/issues/1551#issuecomment-1349918801

Issue #1577, opened 2022-12-13
sdkconfig optimization


nliaudat:
### The Feature

Hi, 

I've done some sdkconfig optimizations : 

```
#better flash mode (may break OTA)
ESPTOOLPY_FLASHMODE_QIO=y

#may trouble very old esp-cam hardware version
ESP32_REV_MIN_3=y
CONFIG_ESP32_REV_MIN_3=y

#set frequency
CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ=240
CONFIG_ESPTOOLPY_FLASHFREQ_80M=y
SPIRAM_SPEED_80M=y

#disable every logging
CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y
CONFIG_COMPILER_OPTIMIZATION_ASSERTION_LEVEL=0
CONFIG_COMPILER_OPTIMIZATION_CHECKS_SILENT=y
CONFIG_FREERTOS_ASSERT_DISABLE=y
CONFIG_HAL_DEFAULT_ASSERTION_LEVEL=0
CONFIG_LOG_DEFAULT_LEVEL_NONE=y
CONFIG_LOG_DEFAULT_LEVEL=0
CONFIG_LOG_MAXIMUM_EQUALS_DEFAULT=y
CONFIG_LWIP_ESP_LWIP_ASSERT=n
CONFIG_OPTIMIZATION_ASSERTIONS_DISABLED=y
CONFIG_OPTIMIZATION_ASSERTION_LEVEL=0

#disable lookup function
CONFIG_ESP_ERR_TO_NAME_LOOKUP=n

#no panic message
ESP_SYSTEM_PANIC_SILENT_REBOOT=y

#disable Access point mode
CONFIG_ESP_WIFI_SOFTAP_SUPPORT=n

#disable ADC calibration (needed for external sensors)
CONFIG_ADC_CAL_EFUSE_TP_ENABLE=n
CONFIG_ADC_CAL_EFUSE_VREF_ENABLE=n
CONFIG_ADC_CAL_LUT_ENABLE=needed

#disable IPV6 (not compatible with : https://github.com/jomjol/AI-on-the-edge-device/pull/1269)
CONFIG_LWIP_IPV6=n

#Newlib format
CONFIG_NEWLIB_NANO_FORMAT=y
```

with that modifications, I get  :

```
[...]

PR #1598, opened 2022-12-15
sdkconfig optimization #1577

nliaudat:
RAM:   [==        ]  19.2% (used 63016 bytes from 327680 bytes)
Flash: [========  ]  78.3% (used 1522617 bytes from 1945600 bytes)


jomjol:
The change is so big, that we should test it in a dedicated path.
1)  This would make 1 device in my home not working anymore:
```
#may trouble very old esp-cam version (work with Chip is ESP32-D0WDQ6 (revision 1))
ESP32_REV_MIN_3=y
CONFIG_ESP32_REV_MIN_3=y 
```

2) I have no idea, what happens, if we disable all debugging and logging
I have no time currently to test and verfy this changes.
3) Currently we do not need addiational space, as there are no big features planned to be implemented. Why should we change?

@caco3 @haverland what is your opinion?

nliaudat:
Hi, 

The disabled logging feature are only esp-cam (aka esp32) logs. All we need as logging is after boot and we do not mind the esp-cam corre

I need more space to implement Matter protocol. :)

You are right with needing a separate dev branch. I mean it must be tested on many devices before release.
I have 1 test device and 2 in production and all pass firmware upgrade, but I seen differences between VS studio compile (use sdkconfig.defauts) and `pio run --target upload` (use skkconfig.esp32cam)

I'll check with the compile version of github if you explain me how to retrieve the binaries or update.zip from the pull request (update__1598_merge_(cb846c7))

Regards

caco3:
@nliaudat You can fetch the auto build from https://github.com/jomjol/AI-on-the-edge-device/actions Just select the right branch, then fetch the update_*.zip. Optionally extract the firmware.bin and just flash that one.

caco3:
I would suggest that we work on the currently open bugs and create another release (13.0.x or 13.1.x.)

Then create a separate branch and play with those optimizations.
Since we have auto build we eg. could have a `rolling-optimized-flash-usage` and periodically merge `rolling` into it.

jomjol:
@caco3 : good idea to have a parallel branch with optimized flash size.

nliaudat:
I confirm the devices are more responsive. I mean due to : 

CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ=240
CONFIG_ESPTOOLPY_FLASHFREQ_80M=y
SPIRAM_SPEED_80M=y

Thanks to create new branch
Regards

caco3:
> CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
> CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ=240

Oh, wow!
@jomjol Is there a reason we do not use it with max speed? (eg. stability)

jomjol:
In the past some device were much more stable with a lower processor speed.

caco3:
maybe we could make it configurable?

Most will want to have a better responsiveness. And if it is not stable (low hardware quality), they can go to slower again.

PR #1604, opened 2022-12-16
code translation from DE to ENG

nliaudat:
Code translation + single place #define in include/define.h


caco3:
@nliaudat Thanks for your effort!

- [ ] I am not fully convinced with `dismantled`, wouldn't `splitted` be a better translation of `zerlegt`? 
- [ ] I would translate `Bereich` with `range`
- [ ] `vorgaenger` should be translated to `predecessor`
- [ ] Can we do separate commits for translations and functional changes? That makes it easier to troubleshoot issues.
- [ ] I think it is better top remove the `defines` on the old places instead of commenting them out.
- [ ] Please do not remove `CONFIG_ESP32_REV_MIN_0` in this MR. It might be ok to remove it but please do it in a separate MR as it is an essential change.

I switched it to draft do it does not get merged by accident.
Also I would suggest to keep this merge block until we did another bugfix release.

nliaudat:
@caco3 

- [done]  I am not fully convinced with dismantled, wouldn't splitted be a better translation of zerlegt?
- [done] I would translate Bereich with range
- [done]   vorgaenger should be translated to predecessor
- [ ]  Can we do separate commits for translations and functional changes? That makes it easier to troubleshoot issues.
= Difficult to split, I've done a lot of job and corrected many things in doing translation
- [done]  I think it is better top remove the defines on the old places instead of commenting them out.
- [done] Please do not remove CONFIG_ESP32_REV_MIN_0 in this MR. It might be ok to remove it but please do it in a separate MR as it is an essential change.

Please look at the final state of my commits. I messed up by mixing everything up a bit, but the final version works and has been tested as an upload on one of my board.

I've also added #define STBI_ONLY_JPEG which save 2% of Flash.

Using one define.h is a good practice. Actually you have  BLINK_GPIO GPIO_NUM_4 in ClassFlowMakeImage.h and BLINK_GPIO GPIO_NUM_33 in all others.

I mean the priority should to clean the code cause I see a lot of copy/paste from many sources without review. 
Many bugs are related to that fact are difficult to undertand cause the code in not clear and with with few comments

nliaudat:
With all the actual improvements, including #1577 : 

CONFIG_COMPILER_OPTIMIZATION_SIZE=y
RAM:   [==        ]  19.2% (used 62976 bytes from 327680 bytes)
Flash: [========  ]  76.9% (used 1496165 bytes from 1945600 bytes)

COMPILER_OPTIMIZATION_PERF=y
RAM:   [==        ]  19.2% (used 63052 bytes from 327680 bytes)
Flash: [========= ]  89.3% (used 1738249 bytes from 1945600 bytes)

caco3:
> With all the actual improvements, including #1577 :

I think we should keep the optimizations separate of the renaming/restructuring!

nliaudat:
sdkconfig.defaults optimizations are not included in that PR

caco3:
@nliaudat  We are very glad about all your contributions, they are great!

What I would like to have:
 - A Pull request just containing translations (and optionally renaming of variables/functions)  =>A lot of changes but low risk
 - A Pull request with restructuring (centralizing the defines) => This is more risky as it is hard to follow, functionality might change/break and we might need to trace it down
 - A Pull request with all the optimizations (flash/RAM) => This is very risky as it could have unexpected side effects.

Separate Pull requests just make it easier to build and test them separately

nliaudat:
how do you translate zw ?

jomjol:
zw = zwischen - I use it for temporary variables.

nliaudat:
The good practice is to name temporary variables as their cast : 
```
i = interger
f = float
d = double
b = boolean
s = strings
etc...
```
or you could be more precise like "path", "file" or whatever
Many thanks for your answer.

jomjol:
your totally right - I'm not a professional, but a autodidact and eager to learn.

nliaudat:
I am also autodidiact and not professional. 

With years, i've learned that in developping  "small is beautiful" is the best way to go. 

-Think of variables footprint = int vs float vs double
-Always mention sources of code as it can be usefull for futher upgrade
-Do not #include everythings 
(I am working on #includes and in main.cpp the followinjg includes are not needed : smarled.h,<string>,"freertos/FreeRTOS.h","freertos/task.h", "freertos/event_groups.h","driver/gpio.h","sdkconfig.h" etc

Good luck in your life as developper !

jomjol:
Thanks! Fortenuatly I do not earn my money with programming :-).

nliaudat:
> Thanks! Fortenuatly I do not earn my money with programming :-).

You are not a developer, but an engineer (the one who brings new solutions). That's a major difference and that's why you have 2800 stars on your git.
Keep it up
Regards

Issue #1574, opened 2022-12-13
Crash if accessing JSON REST API too early after Startup


michaelmosmann:
### The Problem

When http://<IP>/json is called on startup it will lead to a reboot. This does not happen if /capture is called.

### Version

v13.0.7

### Logfile

```shell
[       8] 2022-12-13T17:32:10	<INF>	[MAIN] =================================================
[       8] 2022-12-13T17:32:10	<INF>	[MAIN] ================== Main Started =================
[       8] 2022-12-13T17:32:10	<INF>	[MAIN] =================================================
[      10] 2022-12-13T17:32:12	<INF>	[MAIN] Initialization completed successfully!
[      17] 2022-12-13T18:32:19	<ERR>	[MQTT IF] Init with no Client_ID () or Last Will Topic (). Abort Init!
[      17] 2022-12-13T18:32:19	<ERR>	[MQTT IF] Init failed, skipping all MQTT publishings in this round!
[       4] 2022-12-13T17:33:59	<INF>	[MAIN] =================================================
[       4] 2022-12-13T17:34:00	<INF>	[MAIN] ==================== Startup ====================
[       4] 2022-12-13T17:34:00	<INF>	[MAIN] =================================================
[       4] 2022-12-13T17:34:00	<INF>	[MAIN] Branch: 'HEAD', Revision: 304b9e0, Date/Time: 2022-12-11 22:36, Web UI: Release: v13.0.7 (Commit: 304b9e0) 
[       4] 2022-12-13T17:34:00	<INF>	[MAIN] Reset reason: Via esp_restart
[       8] 2022-12-13T17:34:04	<INF>	[MAIN] =================================================
[       9] 2022-12-13T17:34:05	<INF>	[MAIN] ================== Main Started =================
[       9] 2022-12-13T17:34:05	<INF>	[MAIN] =================================================
[       9] 2022-12-13T17:34:05	<WRN>	[MAIN] Failed to initialize camera module, retrying...
[      16] 2022-12-13T17:34:11	<INF>	[MAIN] Initialization completed successfully!
[       4] 2022-12-13T17:34:17	<INF>	[MAIN] =================================================
[       4] 2022-12-13T17:34:18	<INF>	[MAIN] ==================== Startup ====================
[       4] 2022-12-13T17:34:18	<INF>	[MAIN] =================================================
[...]

PR #1609, opened 2022-12-17
Block REST API calls till resource is ready

Slider0007:
Fix for #1574:

- Block REST API calls until needed resource is ready to use --> prevent unhandled execption and crash of device

@caco3, @jomjol:
Which HTML error code should to be returned in such a case? --> HTTPD_403_FORBIDDEN?
Is it even better to return no error code?


caco3:
I think `503 Service Unavailable` would be the best.

caco3:
Since 503 is not available I agree 403 fits the best, especially with the given error message.

We could implement the 503 by ourselfs, but I think its not worth the time right now.

PR #1639, opened 2022-12-19
#ifndef in *.h + #pragma once

nliaudat:
Hi, 

I've found many not necessary #include

As sample in main.cpp all the commented out works for compile

```
//#include <string>
//#include "freertos/FreeRTOS.h"
//#include "freertos/task.h"
//#include "freertos/event_groups.h"

//#include "driver/gpio.h"
//#include "sdkconfig.h"
//#include "esp_psram.h" // Comming in IDF 5.0, see https://docs.espressif.com/projects/esp-idf/en/v5.0-beta1/esp32/migration-guides/release-5.x/system.html?highlight=esp_psram_get_size
//#include "spiram.h"
#include "esp32/spiram.h"


// SD-Card ////////////////////
//#include "nvs_flash.h"
#include "esp_vfs_fat.h"
//#include "sdmmc_cmd.h"
#include "driver/sdmmc_host.h"
//#include "driver/sdmmc_defs.h"
///////////////////////////////


#include "ClassLogFile.h"

#include "connect_wlan.h"
#include "read_wlanini.h"

#include "server_main.h"
#include "server_tflite.h"
#include "server_file.h"
#include "server_ota.h"
#include "time_sntp.h"
//#include "ClassControllCamera.h"
#include "server_main.h"
#include "server_camera.h"
#ifdef ENABLE_MQTT
    #include "server_mqtt.h"
#endif //ENABLE_MQTT
//#include "Helper.h"
#include "../../include/defines.h"
//#include "server_GPIO.h"
```

I have no utilities like Clang tidy available to cleanup the duplicates includes and for temporary quick fix, I'll propose to add 

```
#pragma once

#ifndef NAME_H
#define NAME_H
...
#endif //NAME_H
```
to all the *.h files
I know `#pragma once` do not work for all compiler (mainly M$), but it do nothing on others


caco3:
The
```
#ifndef NAME_H
#define NAME_H
...
#endif //NAME_H
```

Should be used for all header files. If it is missing, we should add it.

The pragma once I almost never use (but I mostly develop on ANSI C)

caco3:
Looking on https://en.wikipedia.org/wiki/Pragma_once its seams that pragma once can be used as a replacement for the include guard. It has some advantages but is not supported on all compilers.

So i would suggest to stick to one and not use both to keep it tidy

nliaudat:
I've added 

```
#pragma once

#ifndef NAME_H
#define NAME_H
```

in all header with that PR.
But it should be great to clean the cpp #includes

It's time consuming as I remove one, compile, see logs and redo

nliaudat:
#pragma once is use manly in M$ suites, but it do not make bad things on other compilers.

I mean #prama once +#ifndef is not a bad thing.

caco3:
@jomjol  For me the PR is ok as it is

PR #1626, opened 2022-12-18
Code translation

nliaudat:
Discussed in https://github.com/jomjol/AI-on-the-edge-device/pull/1604

Partial, not every function is done

PR #1661, opened 2022-12-21
Reverting changes sdkconfig.defaults

caco3:
Reverting changes of `sdkconfig.defaults` falsly added within https://github.com/jomjol/AI-on-the-edge-device/pull/1626


caco3:
@jomjol @nliaudat @Slider0007 FYI

Issue #1643, opened 2022-12-19
Home Assistant MQTT state_topic path incorrect when using default Number name


DrCoolHands:
### The Problem

I have the meter reader configured to send data to MQTT and to provide HomeAssistant discovery details, however a number of the state_topic fields for HomeAssistant are invalid and include an additional `/default/` path.

From the MQTT Server:

for example:

homeassistant/sensor/aiwatermeter/value/config
```
{
"~": "aiwatermeter",
"unique_id": "aiwatermeter-value",
"object_id": "aiwatermeter_value",
"name": "Value",
"icon": "mdi:gauge",
"state_topic": "~/default/value",
"unit_of_meas": "m³",
"device_class": "water",
"state_class": "total_increasing",
"availability_topic": "~/connection",
"payload_available": "connected",
"payload_not_available": "connection lost",
"device": {
"identifiers": ["aiwatermeter"],
"name": "aiwatermeter",
"model": "Meter Digitizer",
"manufacturer": "AI on the Edge Device",
"sw_version": "v13.0.8",
"configuration_url": "http://192.168.0.100"
}
}
```

The state_topic value is published as '\~/default/value' however that topic does not exist. Instead it should be '~/value'

![image](https://user-images.githubusercontent.com/24321125/208535220-81d5fd9e-e4c2-42be-aa08-3404fb2cc063.png)

This effects the following sensor topics:

- homeassistant/sensor/aiwatermeter/value/config
- homeassistant/sensor/aiwatermeter/raw/config
- homeassistant/sensor/aiwatermeter/error/config
- homeassistant/sensor/aiwatermeter/rate_per_time_unit/config
- homeassistant/sensor/aiwatermeter/rate_per_digitalization_round/config
- homeassistant/sensor/aiwatermeter/timestamp/config
- homeassistant/sensor/aiwatermeter/json/config
- homeassistant/binary_sensor/aiwatermeter/problem/config

The other config topics show the correct state_topic path

Config:
```
[MQTT]
Uri = mqtt://mqtt.local:1883
MainTopic = aiwatermeter
ClientID = watermeter
HomeassistantDiscovery = true
MeterType = water_m3
```

### Version

13.0.8

### Logfile

```shell
[0d00h00m04s] 2022-12-20T03:43:55 <INF> [MAIN] ==================== Startup ====================
[0d00h00m04s] 2022-12-20T03:43:55 <INF> [MAIN] =================================================
[...]

PR #1664, opened 2022-12-21
fix special case where number is named "default" (keep all topics in top level instead of in a sub-group)

caco3:
See https://github.com/jomjol/AI-on-the-edge-device/issues/1643#issuecomment-1361267807

@jomjol can you confirm that the "default" is treated differently?
Is this a legacy thing?
Because the default config seems to use "main" .


jomjol:
yes, can confirm the legacy mode for default. At the beginning there was only one number possible. To handle "old" config files, the default was implemented.

PR #1671, opened 2022-12-22
Implement WLAN 802.11k und 802.11v

jomjol:
The ESP32 support for WLAN 802.11k und 802.11v is implemented. With this a roaming in case of bad RSSI values is possible.
The setting is stored in the `wlan.ini` and can be adjusted within in the Expert Mode of the configuration:
grafik

The parameter range for the RSSI-Threashold can be between -100 and 0. To switch it off set the value to "0".

In `defines.h` is a compiler switch to turn it completly off.

This integrated the feature request from issue #1141


bjoernmayer:
Hey,
doesn't seem to work too well.
I've tried with various thresholds -50, -65, -80. The ESPs keep jumping to one of the farthest AP at my place.
Shortly after reboot, I can see them being connected to the right AP, before jumping to the wrong one shortly after

Issue #1649, opened 2022-12-21
Remove unnecessary null pointer checks


elfring:
### Issue

[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [ClassFlowAlignment::doFlow](https://github.com/jomjol/AI-on-the-edge-device/blob/f6369ff2378cde4f0a36411f1047eed223127fdd/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp#L162 "ClassFlowAlignment::doFlow function")
- [CTfLiteClass](https://github.com/jomjol/AI-on-the-edge-device/blob/f6369ff2378cde4f0a36411f1047eed223127fdd/code/components/jomjol_tfliteclass/CTfLiteClass.cpp#L283 "Destructor for CTfLiteClass")

elfring:
:thought_balloon: I suggest to reconsider another source code place.
https://github.com/jomjol/AI-on-the-edge-device/blob/66eb1e8d9a19376c3e7489b617b4d1555d9f05d7/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp#L213

caco3:
@haverland I think you solved this, right?

haverland:
@elfring: Thanks for you collaboration, but could you make a Pull Request on it?

@caco3: No. @elfring  found more places.

PR #1673, opened 2022-12-22
Remove unnecessary null pointer checks #1649

haverland:


elfring:
:thought_balloon: Why did you mix changes for different issues here?

haverland:
@elfring : comes from back merging. It's only the fix of the title. Look at the file changes.

PR #1698, opened 2022-12-27
macro definition #ifdef ENABLE_SOFTAP

nliaudat:
set macro definition for SoftAP
I mean SoftAP must be enabled only in [initial_esp32_setup.zip](https://github.com/jomjol/AI-on-the-edge-device/releases/download/v13.0.8/initial_esp32_setup.zip)

With SoftAP :
```
RAM:   [==        ]  20.6% (used 67412 bytes from 327680 bytes)
Flash: [==========]  98.8% (used 1921889 bytes from 1945600 bytes)
```

Without SoftAP :
```
RAM:   [==        ]  20.6% (used 67420 bytes from 327680 bytes)
Flash: [==========]  98.4% (used 1914469 bytes from 1945600 bytes)
```

Without SoftAP and CONFIG_ESP_WIFI_SOFTAP_SUPPORT=n
```
RAM:   [==        ]  20.5% (used 67292 bytes from 327680 bytes)
Flash: [==========]  97.0% (used 1886281 bytes from 1945600 bytes)
```


jomjol:
> I mean SoftAP must be enabled only in [initial_esp32_setup.zip](https://github.com/jomjol/AI-on-the-edge-device/releases/download/v13.0.8/initial_esp32_setup.zip)

This will not work. Firstly it is needed in the firmware.bin, which is used by the WebInstaller.
If everything is running right BEFORE the first reboot, it is not needed anymore, therefore also not in the `initial_esp32_setup.zip)`
BUT: if anything goes wrong (e.g. wrong zip-file!), the softAP is needed again.

Therefore currently I suggest to make it a default component, that is not disabled.

@nliaudat : what other changes do we have to make the firmware smaller, but not loose comfort?

jomjol:
Anyway: thanks a lot for making everything configurable. I will merge it, but then enable it by default. Okay?

nliaudat:
It's already enabled in platform.ini. (I never disable anything by default,  just make a way to disable if needed)
`build_flags = -D ENABLE_MQTT -D ENABLE_INFLUXDB -D ENABLE_SOFTAP`

nliaudat:
@jomjol #what other changes do we have to make the firmware smaller, but not loose comfort?

in defines.h :
`#define STBI_ONLY_JPEG // (save 2% of Flash)`
is safe as only jpeg is used. I've tested and it works great.

nliaudat:
https://github.com/jomjol/AI-on-the-edge-device/pull/1706

nliaudat:
> > I mean SoftAP must be enabled only in [initial_esp32_setup.zip](https://github.com/jomjol/AI-on-the-edge-device/releases/download/v13.0.8/initial_esp32_setup.zip)
> 
> This will not work. Firstly it is needed in the firmware.bin, which is used by the WebInstaller. If everything is running right BEFORE the first reboot, it is not needed anymore, therefore also not in the `initial_esp32_setup.zip)` BUT: if anything goes wrong (e.g. wrong zip-file!), the softAP is needed again.

I do not agree : 
1) SoftAP was introduced in https://github.com/jomjol/AI-on-the-edge-device/pull/1692 yesterday and is not use in Webinstaller which push the firmware with serial communication.
2) SoftAP enables only if wlan.ini or config.ini is missing. Such case is true only in initial setup with sdcard not prepared correctly. A remove of sdcard and push the files will fix that situation without the need of SoftAP. 
3) SoftAP has a huge cost of 35kb in flash

I mean as the actuall software takes 98.8% of flash, such improvement must be balances with the cost of futur implementations
Regards

jomjol:
Thanks to your improvements, we now are at 86%. From my point of view we can effort currently the space. It will make the setup for newbees much easier in the future.

PR #1705, opened 2022-12-27
Fetch an available resource after reboot

Slider0007:
Reduce 404 errors at startup phase:

- Fetch an available resource instead of hostname after reboot initiated to avoid a ton of 404 errors at startup phase
- Lower frequency of "pings" -> only every 5s

### DIFF
--- a/code/components/jomjol_fileserver_ota/server_ota.cpp
+++ b/code/components/jomjol_fileserver_ota/server_ota.cpp
@@ -628,7 +628,7 @@ esp_err_t handler_reboot(httpd_req_t *req)
 
     LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "handler_reboot");
     ESP_LOGI(TAG, "!!! System will restart within 5 sec!!!");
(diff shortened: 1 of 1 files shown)


caco3:
@Slider0007 After this change I now always get a 404 page shown after a reboot.
I think it is because the device response, but the sdcard handler is not ready yet. Can you revert or improve it again?

Slider0007:
@caco3 : I'll have a look or revert it just back. Which brower are you using?

Slider0007:
@caco3 : Can you give this version a try. 

fetch hostname + updated overview: https://github.com/jomjol/AI-on-the-edge-device/actions/runs/3800836867
fetch index,html + updated overview: https://github.com/jomjol/AI-on-the-edge-device/actions/runs/3800588373

Both works for me from GUI point of view. But fetching index.html avoids 404 erros on serial console. 
I have only to refresh the overview manually when the resources like flow, etc. are not yet availabale.

caco3:
I use Firefox

 - fetch hostname + updated overview
   This works (no 404), but the progress bar using the dots is kind of useless with a 5s interval!
- fetch index,html + updated overview
  That one works as well, but the UI misses the logo and photo after the reload:
  grafik
 and also the progress bar is not as useful anymore.


If you want to stick with 5s, we could add a 2nd timer for the progress bar.

PR #1706, opened 2022-12-27
Safe optimizations

nliaudat:
No IPV6, log level set to error in bootloader, STBI_ONLY_JPEG and many small optimizations :
I mean it's safe without any cpu, spiram, flash and board version optimization.
I cannot test actually on real board

Actual : 
RAM:   [==        ]  20.6% (used 67428 bytes from 327680 bytes)
Flash: [==========]  98.8% (used 1921797 bytes from 1945600 bytes)

Opt : 
RAM:   [==        ]  19.9% (used 65100 bytes from 327680 bytes)
Flash: [========= ]  85.7% (used 1666969 bytes from 1945600 bytes)


caco3:
I reverted the `#define STBI_ONLY_JPEG` as it breaks the alignment mark generation, see https://github.com/jomjol/AI-on-the-edge-device/issues/1721

Issue #1715, opened 2022-12-29
Regression on restart handling


caco3:
### The Problem

On rolling I see 2 issues when triggering a soft reset:

1. Crash:
  ```
  ␛[0;32mI (70859) OTA: Trigger reboot due to firmware update.␛[0m
  ␛[0;32mI (70869) OTA: Reboot triggered by Software (5s).␛[0m
  ␛[0;33mW (70889) OTA: Reboot in 5sec␛[0m
  Guru Meditation Error: Core  0 panic'ed (IllegalInstruction). Exception was unhandled.
  ```

2. Restart ignored:
  ```
  [0d04h02m03s] 2022-12-29T15:53:19 <INF> [OTA] Reboot triggered by Software (5s).
  [0d04h02m03s] 2022-12-29T15:53:19 <WRN> [OTA] Reboot in 5sec
  [0d04h02m04s] 2022-12-29T15:53:20 <INF> [POSTPROC] main: Raw: 521.9356, Value: 521.9356, Status: no error
  [0d04h02m04s] 2022-12-29T15:53:20 <INF> [TFLITE SERVER] Round #377 completed
  [0d04h02m04s] 2022-12-29T15:53:20 <INF> [TFLITE SERVER] Round #378 started
  [0d04h02m09s] 2022-12-29T15:53:25 <INF> [CAM] Using /sdcard/demo/520.9787.jpg as demo image
  [0d04h02m09s] 2022-12-29T15:53:25 <INF> [CAM] read 24006 bytes
  [0d04h02m36s] 2022-12-29T15:53:51 <INF> [OTA] Reboot triggered by Software (5s).
  [0d04h02m36s] 2022-12-29T15:53:51 <WRN> [OTA] Reboot in 5sec
  [0d04h02m47s] 2022-12-29T15:54:03 <INF> [POSTPROC] main: Raw: 521.9787, Value: 521.9787, Status: no error
  [0d04h02m47s] 2022-12-29T15:54:03 <INF> [TFLITE SERVER] Round #378 completed
  ``


### Version

rolling

### Logfile

```shell
.
```

Slider0007:
- Restart ignored --> This is already fixed in #1713 
- OTA issue --> Should be addressed with PR #1725

PR #1725, opened 2022-12-29
Fix regression of reboot handler / reboot waiting script

Slider0007:
Clean up the reboot handler topics comming up after my last changes (sorry guys for messing up).

- Fix regression after OTA reboot mentioned here #1715
- Restart ignored (mentioned also in #1715) was already fixesd with #1713
- Reboot waiting script (#1705):
   --> Revert ping time of 5s and set it back 1s
   --> Changed ping resource to ensure more reliable page reloading
- Optimization of overview.html (in addition to #1711)

@caco3 : Please test reboot behaviour, especially in terms of page reloading of overview after reboot


caco3:
Thanks @Slider0007 for your work!

It looks good to me.

PR #1736, opened 2022-12-31
#1649 unnecessary np check

haverland:


elfring:
Thanks for your deletion of another redundant pointer check.

PR #1752, opened 2023-01-03
ClassFlowAlignment - Allocate only one ImageTMP

Slider0007:
Allocate only one ImageTMP instead 5x

@jomjol: Do I have something overseen here?


jomjol:
> Allocate only one ImageTMP instead 5x
> 
> @jomjol: Do I have something overseen here?

Well, it is not, that it makes 5 images with this parameters, but it tries it 5 times, if there is not enough free memory. 
That is coming from older times, where the malloc failed from time to time due to missing heap size. 

I checked the code and this parameter is not usefull anymore, therefore I will remove this function everywhere.

jomjol:
Functions cleaned in newest rolling.

Slider0007:
@jomjol : Thanks for your explanation and taking care of it. I misinterpreted somehow. I think it was too late yesterday 😉

Issue #1747, opened 2023-01-02
Cache html pages


caco3:
### The Feature

Add "Cache-Control" header to reduce load on webserver

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control?retiredLocale=de
- https://github.com/esp8266/Arduino/issues/999

PR #1755, opened 2023-01-03
cache static files

caco3:
implements https://github.com/jomjol/AI-on-the-edge-device/issues/1747

### DIFF
--- a/code/components/jomjol_fileserver_ota/server_help.cpp
+++ b/code/components/jomjol_fileserver_ota/server_help.cpp
@@ -28,6 +28,15 @@ static const char *TAG = "SERVER HELP";
 
 char scratch[SERVER_HELPER_SCRATCH_BUFSIZE];
 
+
+bool endsWith(std::string const &str, std::string const &suffix) {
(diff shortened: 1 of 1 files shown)


Slider0007:
@caco3, @jomjol, @haverland : We have to keep in mind that files are cached now and html changes can be "hidden" in first instance because still "old" cached html files are shown.

I figured out with Ctrl + F5 it's possible to force a complete refresh and ignore cached files.
Is there any other possibility to have a comlete refresh e.g. after OTA upgrade?

jomjol:
One can add a kind of hash string at the end of the download file in the following way:

caco3:
> I figured out with Ctrl + F5 it's possible to force a complete refresh and ignore cached files.

I am sorry, i should have mentioned that but thought that's common knowledge

> One can add a kind of hash string at the end of the download file in the following way: <script type="text/javascript" src="common.js?v=12392823"></script>

Yes, I think thats the way to go. I will create a PR for that which uses the hash of the build

PR #1767, opened 2023-01-04
Add testing environnements in platformio

nliaudat:
1)Add fine tune for testing : 
```
[env:esp32cam-testing]
build_flags = 
	-D ENABLE_MQTT -D ENABLE_INFLUXDB -D ENABLE_SOFTAP
    -D CONFIG_ESP_TASK_WDT
    -D CONFIG_SPIRAM
    -D CONFIG_ESP_TASK_WDT_TIMEOUT_S ; fix for CONFIG_ESP_INT_WDT_TIMEOUT_MS
lib_ldf_mode = deep+
platform = platformio/espressif32 @ 5.2.0
platform_packages = 
    platformio/tool-esptoolpy@^1.40400.0
    espressif/toolchain-esp32ulp @ 2.35.0-20220830
  
```

2) More boards to test :
Add testing support for several boards as esp32c3, esp32s2, rp2040

3) Usage : 
You can invoke differents environnements :
```
platformio run --environment esp32cam-testing
platformio run --environment esp32c3-idf-testing
platformio run --environment esp32s2-idf-testing
platformio run --environment rp2040-pico-arduino-testing
```

4) Notes : 
Only compile actually with esp32cam-testing, but do not infer with classic "esp32cam"
(You may need to delete your .platformio\packages and clean all cache, before compiling other envs)

5) Finally : 
AI-Thinker-esp32cam board is end of life https://github.com/jomjol/AI-on-the-edge-device/discussions/1732, we need to prepare for future ...

### DIFF
--- a/code/platformio.ini
+++ b/code/platformio.ini
@@ -79,6 +79,7 @@ board = esp32cam
 framework = espidf
 build_flags = 
 ;Add macro definition ENABLE_MQTT, ENABLE_INFLUXDB, DEBUG_DETAIL_ON
+;if ENABLE_SOFTAP = disabled, set CONFIG_ESP_WIFI_SOFTAP_SUPPORT=n in sdkconfig.defaults to save 28k of flash
 	-D ENABLE_MQTT -D ENABLE_INFLUXDB -D ENABLE_SOFTAP
 	${common:esp32-idf.build_flags}
 	${flags:runtime.build_flags}
@@ -93,6 +94,7 @@ board = node32s
 board_build.flash_mode = qio
 build_flags = 
 ;Add macro definition ENABLE_MQTT, ENABLE_INFLUXDB, DEBUG_DETAIL_ON
+;if ENABLE_SOFTAP = disabled, set CONFIG_ESP_WIFI_SOFTAP_SUPPORT=n in sdkconfig.defaults to save 28k of flash
 	-D ENABLE_MQTT -D ENABLE_INFLUXDB -D ENABLE_SOFTAP
 	${common:esp32-idf.build_flags}
 	${flags:clangtidy.build_flags}
--- a/code/platformio.ini
+++ b/code/platformio.ini
@@ -8,29 +8,144 @@
 ; Please visit documentation for the other options and examples
 ; https://docs.platformio.org/page/projectconf.html
 
-
 [platformio]
 src_dir = main
+default_envs = esp32cam
+
+[common:idf]
+build_flags = 
+	-DUSE_ESP_IDF
+lib_deps =
+
+[common:arduino]
+extends = common
+lib_deps =
+build_flags =
+    ${common.build_flags}
+    -DUSE_ARDUINO
+
+[common:esp32-idf]
+extends = common:idf
+platform = platformio/espressif32 @ 5.2.0
+platform_packages = 
+	;platformio/framework-espidf @ ~3.40402.0
+framework = espidf
+lib_deps = 
+	${common:idf.lib_deps}
+build_flags = 
+	${common:idf.build_flags}
+	-Wno-nonnull-compare
+	-DUSE_ESP32
+	-DUSE_ESP32_FRAMEWORK_ESP_IDF
+
+
+; These are common settings for the RP2040 using Arduino.
+[common:rp2040-arduino]
+extends = common:arduino
+board_build.core = earlephilhower
+board_build.filesystem_size = 0.5m
+
+platform = https://github.com/maxgerhardt/platform-raspberrypi.git
+platform_packages =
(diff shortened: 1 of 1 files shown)


jomjol:
This is a lot of change!

To be honest I don't understand most of them - which means I could not change them if something goes wrong. I try to keep this file as simple as possible, because I experiences myslelft, that it is a major barrier for newbees to platform.io.

The more debugging and compiler flags we add, the less I and others understand, what they trigger and why they are need.

@nliaudat : can you explain, what the additional chapters are and why we need them:

```
[common:idf]
[common:arduino]
[common:esp32-idf]
[common:rp2040-arduino]
[flags:runtime]
[flags:clangtidy]
[env:esp32cam]
[env:esp32cam-testing]
[env:esp32c3-idf-testing]
[env:esp32s2-idf-testing]
[env:rp2040-pico-arduino-testing]
```

For example I don't get, why we need an arduino or rp2040-aduiono, ..., the software will not run within it.

If this is need for testing or comparable, I would suggest to make a branch for this.

nliaudat:
## Platformio 

Platformio works like this :
[Common [env]](https://docs.platformio.org/en/latest/projectconf/sections/env/index.html#common-env)
[Working [env:NAME]](https://docs.platformio.org/en/latest/projectconf/sections/env/index.html#working-env-name)
[Options](https://docs.platformio.org/en/latest/projectconf/sections/env/index.html#options)

The environnements are hierarchical and that's why you'll find [common:idf] =>[common:esp32-idf] =>[env:esp32cam]

[common:idf]
build_flags = -DUSE_ESP_IDF

=>
[common:esp32-idf]
_**extends = common:idf**_
platform = platformio/espressif32 @ 5.2.0

=>
[env:esp32cam]
**_extends = common:esp32-idf_**
board = esp32cam


## Build flags
For the build flags, the runtime version do only the minimum. As example, the "Wno-unused-but-set-variable" flag is quite self explanatory. The clangtidy runtime flags will run fast all the compiler checks available and is used to find mistakes in the dev. The runtime flags compiles faster, but are subject to memory leaks, etc

## Testing - Dev environnement

The previously discussed build flags are very important before making a release. 
But in the testing env, you'll find more checks like 
[lib_ldf_mode ](https://docs.platformio.org/en/stable/librarymanager/ldf.html)= deep+ 

And you can easily test next platform release, perhaps as setting
`platform = platformio/espressif32 @ 5.3.0`
or perhaps platform package 
`espressif/toolchain-esp32ulp @ 2.35.0-20220830`

## Other Boards + rp2040 / arduino
The Ai-Thinker espcam is end of life.
Espressif has released AI vision boards like ESP32-EYE and ESP32-S3-EYE and they need soon to be supported.
Raspberry Pi rp2040 is also a good alternative as support tflite and camera (with hardware module). That's the why of a dev env with arduino

jomjol:
Okay, understood a bit more.

Regarding the other hardware versions:
* I don't see a quick switch as all of them have not direct SD-slot integrated - resulting in major changes or even decrease of fucntionality
* Therefore we should take them out of the master branch and in case there is some interest, make a dedicated branch for them.

This reduction should make the `platform.ini` much smaller with a better overview.

@caco3 , @haverland : what's your opinion about this?

nliaudat:
I understand your fear, but please accept the [env:esp32cam-testing]

It can solve problems like : 

```
components/jomjol_fileserver_ota/server_ota.cpp:132:9: warning: unused variable '_i' [-Wunused-variable]
     int _i = configMINIMAL_STACK_SIZE;
         ^~
```
or
```
components/jomjol_wlan/connect_wlan.cpp:82:21: warning: 'sta_netif' defined but not used [-Wunused-variable]
 static esp_netif_t *sta_netif = NULL;
```

jomjol:
I have integrated the changes for esp32cam directly into the current rolling.
* 1:1 copy from this platformio.ini, but removed the sections for other boards.

Therfore this pull request is "merged" with modifcatins and I close this.

haverland:
> Okay, understood a bit more.
> 
> Regarding the other hardware versions:
> 
> * I don't see a quick switch as all of them have not direct SD-slot integrated - resulting in major changes or even decrease of fucntionality
> * Therefore we should take them out of the master branch and in case there is some interest, make a dedicated branch for them.
> 
> This reduction should make the `platform.ini` much smaller with a better overview.
> 
> @caco3 , @haverland : what's your opinion about this?

Think is a good idea. The "testing" confused me a bit. Because it's not for running the tests. Better "-dev"?

I'm with you with the alternativ hardware. Currently I see no need to support more hardware.
But if it's in the master branch, it suggests we support the hardware. 
If you plan to support it we should - as long as we not support it -  comment out it and add a comment that is not supported hardware currently. All in the rolling brach can (and will sometimes) in the master branch.

PR #1784, opened 2023-01-07
Add option to disable brownout detector

nliaudat:
It's sometime usefull to disable brownout detector (low voltage)
Add debug option in defines.h : #define DISABLE_BROWNOUT_DETECTOR

### DIFF
--- a/code/include/defines.h
+++ b/code/include/defines.h
@@ -16,7 +16,9 @@
     */
     // server_tflite.cpp
     //#define TASK_ANALYSIS_ON
-    
+
+    // ######## debug options : 
+    //#define DISABLE_BROWNOUT_DETECTOR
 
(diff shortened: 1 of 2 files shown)


caco3:
Are you sure this is wise?
Brownout usually means the Power Supply to the CPU is too low to function correctly. It could hang or do something else odd. And without the reset you have to go and reset it manually!

nliaudat:
That's a debug option. Not enabled but you can use it if needed

Actually, I have 2 board with bad soldering and the run only if brownout is disabled. 
They are burn out I mean, but it's good for debugging. 
Regards

caco3:
Ah, right.
Sry, I only read the description but didn't review the code.

Looks ok for me!

PR #1809, opened 2023-01-11
Store preprocessed image with ROI to RAM

### LINKED ISSUE #1799: Alg_roi.jpg store into memory directly as JPG
### The Feature

Store alg_roi.jpg into memory directly as JPG
- Reduce memory usage 
- Faster refresh

Slider0007:
I created a version to be tested with these changes to implement and close #1799:

- Image (alg_roi) is going to be saved / recalled from RAM as JPG preprocessed with ROI included
--> reduced RAM usage by ca. 800KB because no additional CImage Object is needed for ROI processing (ImageTMP will be used instead)
--> faster loading on web interface
--> model dig-cont_0600 can be used without RAM issues (errors or crashes) anymore 
- Image with included ROIs will be shown only if alignment is finished
- Till alginment is finished the raw image will be shown instead
- Show still the real processing of the image during take image phase (default, configurable) If disabled the update of the image will be done first when image is already taken.
- At the beginning of Take image, old raw image gets deleted to not show image from previous round (black image)

- removed zwischenspeicher array out of CaptureToBasisImage (@jomjol: Was there any special reason or issue camera framebuffer was saved first to zwischenspeicher and only then to CImage object?)

@caco3, @jomjol, @haverland: I successfully tested with two different devices, but I recommend more testing...


jomjol:
I've tested on 4 more devices at home. It works like a charm. I suggest to merge!

Great work!!!

caco3:
I now also tested it on 2 devices, did not find any issue!

Issue #1827, opened 2023-01-13
remove automatic restart in case of CaptureToBasisImage error


caco3:
### The Feature

Retry (now or in next round) in case of below error:
```
E (34912) CAM: is not working anymore (CCamera::CaptureToBasisImage) - most probably caused by a hardware problem (instablility, ...). System will reboot.
I (34942) OTA: Reboot triggered by Software (5s).
W (34942) OTA: Reboot in 5sec
```

An automated reboot always should be avoided.

Maybe we could add a retry and only if the issue persists do a reboot.

stefan123t:
I wonder if the camera or its framebuffer has to / can be re-initialized:
Maybe some PSRAM cleanup and polish would allow it to get ready for the next capture ?
We could even run a kind of memory test in case this error occurs, which would be better than simply resetting the device.

jomjol:
In principle your totally right and I assume, the cleanup and initialization of the camera is basically the part of the reboot to remove the error.
I just handled it that way, because I'm not deep in the camera subsystem and that was the easiest way to get to a definite clean state. Could be improved.

caco3:
Lets see what happens when the automatic reboot is disabled.
- If it works again on the next round -> problem solved
- If it stays broken -> We need to enable the reboot again or find a way to fix the issue.
I made a PR for this: https://github.com/jomjol/AI-on-the-edge-device/pull/1837

@stefan123t Please test and let us know.

 You can get it from https://github.com/jomjol/AI-on-the-edge-device/actions/workflows/build.yaml?query=branch%3Adisable-automatic-restarts

We would like to create a new release soon, so if you can give us a feedback we can add this change into it.

@jomjol I think it would be good to have this change in the next release.

stefan123t:
@caco3 I have seen your update only now.
I am in the process of flashing the AI-on-the-edge-device__manual-setup__1837_merge_\(6460f0d\).zip from the build action you mentioned above. I will need to update the SD-Card and do some trials...

stefan123t:
I did a couple of smoke tests but could not reproduce the erroneous behaviour.
It worked without the known problem this time. 
I may have to revert it to the erroneous version 13.0.8 or rolling to do a double check. 

While stepping through the config wizard I have seen the following dialogs multiple times:
* Configuration got updated. Please reboot to activate changes!
[...]

PR #1837, opened 2023-01-14
disable the automatic reboots in case the camera is no longer working or an empty image got loaded.

caco3:
Hopefully it works again in the next round. If not, we need to add the reboot again.

Removed automatic restarts for:
 - CaptureToBasisImage: Capture Failed
 - Image with size 0 loaded


caco3:
Waiting for feedback from @stefan123t

stefan123t:
@caco3 provided feedback in #1827

caco3:
@stefan123t Thank you for your feedback!

> I did a couple of smoke tests but could not reproduce the erroneous behaviour.
It worked without the known problem this time.
I may have to revert it to the erroneous version 13.0.8 or rolling to do a double check.

As long as we don't know that the changes of this PR are an improvement, I do not want to merge it.

PR #1839, opened 2023-01-14
Improve data logging on errors

caco3:
- There was a redundant call to `WriteToData()` in case of "Rate to high".
  This lead to the `value` to show the too-high value instead of using the `prevalue`

- Also in case of an error, the time did not get updated
  @jomjol Was there a reason for not updating the time, see my changes in https://github.com/jomjol/AI-on-the-edge-device/pull/1839/commits/c918254405fc6699429369771227a003db2836fe

Before:
```
2023-01-14T23:20:18+0100,main,530.0710,530.0710,530.0710,0.062842,0.0597,no error,5.0,2.9,0.0,0.7,7.0,0.8,0.6
### Next line: rate to high but time is not updated
2023-01-14T23:20:18+0100,main,531.8223,,530.0710,,0.0597,Rate too high - Read: 531.8223 - Pre: 530.0710 - Rate: 1.7513,5.0,3.0,1.7,8.2,2.2,2.1,3.2
2023-01-14T23:22:19+0100,main,530.0742,530.0742,530.0742,0.001587,0.0032,no error,5.0,2.9,0.0,0.7,7.3,4.5,2.8
2023-01-14T23:23:19+0100,main,530.1225,530.1225,530.1225,0.048300,0.0483,no error,5.0,2.9,0.0,1.2,2.3,2.3,5.1
```

New:
```
2023-01-14T23:28:21+0100,main,530.0113,,530.21360,,,Neg. Rate - Read:  - Raw: 530.0113 - Pre: 530.2136 ,5.0,2.9,9.9,0.2,1.1,1.2,3.6
2023-01-14T23:29:19+0100,main,530.0710,,530.21360,,,Neg. Rate - Read:  - Raw: 530.0710 - Pre: 530.2136 ,5.0,2.9,0.0,0.7,7.0,0.8,0.6
2023-01-14T23:30:19+0100,main,531.8223,,530.21360,,,Rate too high - Read: 531.8223 - Pre: 530.2136 - Rate: 1.6087,5.0,3.0,1.7,8.2,2.2,2.1,3.2
2023-01-14T23:31:19+0100,main,530.0742,,530.21360,,,Neg. Rate - Read:  - Raw: 530.0742 - Pre: 530.2136 ,5.0,2.9,0.0,0.7,7.3,4.5,2.8
2023-01-14T23:32:19+0100,main,530.1225,,530.21360,,,Neg. Rate - Read:  - Raw: 530.1225 - Pre: 530.2136 ,5.0,2.9,0.0,1.2,2.3,2.3,5.1
```

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
@@ -734,8 +734,13 @@ bool ClassFlowPostProcessing::doFlow(string zwtime)
         NUMBERS[j]->ReturnValue = "";
         NUMBERS[j]->ErrorMessageText = "";
         NUMBERS[j]->Value = -1;
-        /* TODO to be discussed, see https://github.com/jomjol/AI-on-the-edge-device/issues/1617 */
-//        NUMBERS[j]->lastvalue = imagetime;    // must only be set in case of good value !!! --> move to the end
+
+        /* calculate time difference BEFORE we overwrite the 'lastvalue' */
+        double difference = difftime(imagetime, NUMBERS[j]->lastvalue);      // in seconds
+
+        /* TODO:
+         * We could call `NUMBERS[j]->lastvalue = imagetime;` here and remove all other such calls further down.
+         * But we should check nothing breaks! */
 
         UpdateNachkommaDecimalShift();
 
@@ -866,7 +871,6 @@ bool ClassFlowPostProcessing::doFlow(string zwtime)
             ESP_LOGD(TAG, "After AllowNegativeRates: Value %f", NUMBERS[j]->Value);
         #endif
 
-        double difference = difftime(imagetime, NUMBERS[j]->lastvalue);      // in seconds
         difference /= 60;  
         NUMBERS[j]->FlowRateAct = (NUMBERS[j]->Value - NUMBERS[j]->PreValue) / difference;
         NUMBERS[j]->ReturnRateValue =  to_string(NUMBERS[j]->FlowRateAct);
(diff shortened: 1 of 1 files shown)


jomjol:
> * Also in case of an error, the time did not get updated
>       @jomjol Was there a reason for not updating the time, see my changes in [c918254](https://github.com/jomjol/AI-on-the-edge-device/commit/c918254405fc6699429369771227a003db2836fe)

YES, there is clear reason. Please have a look on the code a little bit further down in the routine:
```
        double difference = difftime(imagetime, NUMBERS[j]->lastvalue);      // in seconds
        difference /= 60;  
        NUMBERS[j]->FlowRateAct = (NUMBERS[j]->Value - NUMBERS[j]->PreValue) / difference;
        NUMBERS[j]->ReturnRateValue =  to_string(NUMBERS[j]->FlowRateAct);
```
The differenc will allway be a zero, if you update the imagetime before that calculation. This does only matter, if you use the "RateChange"-Mode (which I do). If you use the absolute change, it will not be a problem.

I have changes this back several time now. We must not update the time before calculating the difference for the rate change.

caco3:
@jomjol Thanks for the explanation!

This would would be worked around if we would move `double difference = difftime(imagetime, NUMBERS[j]->lastvalue);` from line 869 up to the top of the `for` loop (eg. line 739), right?

As far as I can see, we anyway call `NUMBERS[j]->lastvalue = imagetime;` at line 904 for all ROIs, so then we also could call it in both error cases.

For cleanliness, we then additionally could add a `NUMBERS[j]->lastvalue = imagetime;` at line 740 and remove all following `NUMBERS[j]->lastvalue = imagetime;` lines.
(This again only if no function gets called in-between which depends on the old `NUMBERS[j]->lastvalue`).

jomjol:
That is right. I could do that on the weekend earliest - or you change that.

caco3:
I just did it myself in the latest commit.

PR #1841, opened 2023-01-14
enhance development environment

nliaudat:
Add easy dev environment in platformio. 

## Note : 
_Do not change anything in main [env:esp32cam] except ${flags:runtime.build_flags} instead of clangtidy. The release compilation will get less verbose for warnings._

## Usage :
As sample, to set the full debug in an existing env, you just have to set 
extends = env:esp32cam-dev, **_esp32cam-debug_**

You can chain options : 
extends = env:esp32cam-dev, env:esp32cam-board-rev3, env:esp32cam-cpu-freq-240 , esp32cam-debug 
If multiple items specified in the extends field then only values from the latter one will be used in the final configuration

Platformio.ini is now well commented

## CMakeLists.txt changes :
All the new env have custom sdkconfig.[pio env].defaults like in sdkconfig.esp32cam-dev-task-analysis.defaults
That sdkconfig are added to sdkconfig.defaults trough a new CMakeLists.txt

update cmake_minimum_required to 3.16.0


caco3:
@nliaudat Does this mean the release has no DEBUG log on USB anymore?
On one side this would speed it up a bit, but makes debugging harder.
If so, we should build a 2nd firmware where the DEBUG logs are still enabled, but this would mean we have to test 2 firmwares! Also people with issues would have to re-install first.

nliaudat:
> @nliaudat Does this mean the release has no DEBUG log on USB anymore? 

No, only the compilation logs in release mode is less verbose. It do not check all clangtidy code cleaning. 
It do not change anything for the software, only the compilation outputs.

caco3:
> No, only the compilation logs in release mode is less verbose. It do not check all clangtidy code cleaning.
It do not change anything for the software, only the compilation outputs.

Thanks for the explanation!
Then I see no issue with this change.

Issue #1857, opened 2023-01-16
Show alternative images for states "Initialization" and "Take Image"


Slider0007:
### The Feature

Follow up to this discussion https://github.com/jomjol/AI-on-the-edge-device/pull/1809 about alternative images for the state "Flow not started" and state "Take image" due to the fact that in these states right now no, black or only partly rendered image is shown.

#########################
Providing an alternative image during "Take Image" step would be an option, also to avoid the black screen at the beginning of the process.

What is also missing is the visualization when flow is not started yet. Also for this we could show an indication image which is loaded from SD.

Possible steps:
1. Device / webserver started, but flow not started (yet) --> Image with indication "flow not started"
2. Flow started and step "Take image" --> Image with indication of taking image
3. Flow started and step "Alignemnt" --> serve alg (like already coded + ?alignment icon?)
4. and the follwoing steps --> serve alg_roi (like already coded)

_Originally posted by @Slider0007 in https://github.com/jomjol/AI-on-the-edge-device/issues/1799#issuecomment-1382484857_

Slider0007:
Implemented with #1858

PR #1858, opened 2023-01-16
Show alternative images for states "Initialization" and "Take Image"

Slider0007:
Implements feature request #1857

-------------------
Follow up to this discussion https://github.com/jomjol/AI-on-the-edge-device/pull/1809 about alternative images for the state "Flow not started" and state "Take image" due to the fact that in these states right now no, black or only partly rendered image is shown.

- Add new state "Initialization (delayed)" and show an image --> State after a panic restart
- Add new state "Initialization" and show an image --> State right before "Take image"
- State "Take image": Show an static alternative image to avoid black or partly rendered image
- Overview.html: Renamed "Status" --> "State" and switched position with "Last page refresh"

Thanks to @caco3 for the updated images.


code/components/jomjol_flowcontroll/ClassFlowControll.cpp
```diff
@@ -195,7 +189,7 @@ void ClassFlowControll::SetInitialParameter(void)
     flowpostprocessing = NULL;
     disabled = false;
     aktRunNr = 0;
-    aktstatus = "Booting ...";
+    aktstatus = "Flow not started";
```
caco3, line 192: is it `not started` or could we say `not yet started`?

caco3:
It works as expected!

The pictures look a bit pixelated. Did you scale them up or compress too much?

Slider0007:
The images are only 640x480, therefore they all look not so good on modern screen ;-) They are not compressed but the source was an icon which might be not the best one. The images can be changed against better looking ones.

Do somebody else have a look to find better looking ones? Otherwise I try to create better looking ones. Is this sytle to follow? What you guys think?

In terms the naming of state "flow not started":
"Flow not yet started" would be an option. Thinking the other direction, maybe "Flow initializing..." could also be an option?

To be complete what about the scenario of postponed start due to a panic restart? An additional state could also indicate this: e.g. "Flow start postponed"

caco3:
I will create some pictures

caco3:
What do you think about my images?

grafik

caco3:
About the term "flow":
I know, internally we use it heavenly, but on the UI or config we instead use "round".
I think it would be good to use the same words for all user interactions.

caco3:
> To be complete what about the scenario of postponed start due to a panic restart? An additional state could also indicate this: e.g. "Flow start postponed"

Yes, that might be useful.
And should we also add a new image for this?

code/components/jomjol_flowcontroll/ClassFlowControll.cpp
Slider0007: states renamed

Slider0007:
Updated version commited. Description updated.

caco3:
As is now, one of the status we send over MQTT is "Sending MQTT":
grafik

grafik

See https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_flowcontroll/ClassFlowControll.cpp#L88

I would suggest to disable sending this state over MQTT.

Slider0007:
> I would suggest to disable sending this state over MQTT.

Just for my understanding, what's your background thinking to remove this?
If this would be disabled what about the state "Sending InfluxDB"?
Maybe it could be combined using a more generic text e.g. "Publish results" for both states or get rid of both states?

caco3:
> > I would suggest to disable sending this state over MQTT.
> 
> Just for my understanding, what's your background thinking to remove this? If this would be disabled what about the state "Sending InfluxDB"? Maybe it could be combined using a more generic text e.g. "Publish results" for both states or get rid of both states?

Its kind of odd that there is a MQTT message which states "sending MQTT message".
We actually are sending at other times as well.

Sending InfluxDB is something else for me as it might be useful to see in MQTT when this happens.
So I just would disabling the MQTT status sending for that step.

But if you disagree it is ok, it doesn't really hurt me :)

Slider0007:
> > > I would suggest to disable sending this state over MQTT.
> > 
> > 
> > Just for my understanding, what's your background thinking to remove this? If this would be disabled what about the state "Sending InfluxDB"? Maybe it could be combined using a more generic text e.g. "Publish results" for both states or get rid of both states?
> 
> Its kind of odd that there is a MQTT message which states "sending MQTT message". We actually are sending at other times as well.
> 
> Sending InfluxDB is something else for me as it might be useful to see in MQTT when this happens. So I just would disabling the MQTT status sending for that step.
> 
> But if you disagree it is ok, it doesn't really hurt me :)

@caco3 : I'm very open minded to this and do not have any preference. I would suggest to move this discussion to a separate topic because it's not directly linked to this PR if this OK for you.

@caco3, @jomjol: From my point of view this PR would be ready for merge if everybody is happy with the implementation.

caco3:
For me that MR also is good.
I would suggest to merge it and then create a new release

caco3 (approved):

PR #1861, opened 2023-01-16
Heap tracing

nliaudat:
Memory leak tracing
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/heap_debug.html#heap-information

need CONFIG_HEAP_TRACING_STANDALONE=y or 
#define CONFIG_HEAP_TRACING_STANDALONE

**all setups are predefined in [env:esp32cam-dev-task-analysis]**

Actually 2 heap tracing are set in main.cpp

        HEAP_TRACING_MAIN_WIFI
        HEAP_TRACING_MAIN_START

Output is like

```
16 bytes (@ 0x3ffd1914) allocated CPU 1 ccount 0x419d1cf8 caller 0x4010a09b:0x4010a0c6
16 bytes (@ 0x3ffd18cc) allocated CPU 1 ccount 0x419d5680 caller 0x4010a09b:0x4010a0c6
78141 bytes 'leaked' in trace (96 allocations)
total allocations 5598 total frees 5447
```
Adresses can be decoded with [idf monitor](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-monitor.html)

```
=> give the code line
0x400d27c1: blink_task /blink.c:52
```

_The actual main.cpp heap trace is not optimal as the trace function should better be set at start/end of function()
This is more a sample setup_


Slider0007:
This sounds quite interesting. Does this also work somehow with platformio or only idf tools?
In your example does it mean that we have a leakage in this area (difference allocations / frees) or is it kind of misleading because the bytes are still in use somewhere?

```16 bytes (@ 0x3ffd1914) allocated CPU 1 ccount 0x419d1cf8 caller 0x4010a09b:0x4010a0c6
16 bytes (@ 0x3ffd18cc) allocated CPU 1 ccount 0x419d5680 caller 0x4010a09b:0x4010a0c6
78141 bytes 'leaked' in trace (96 allocations)`
total allocations 5598 total frees 5447`

**Comment by nliaudat** [comment]
Hi, 

> Does this also work somehow with platformio or only idf tools?

It'works standalone as it logs to the printf board, but if you want decod addresses space to function, you need idf monitor. 

> have a leakage in this area (difference allocations / frees) 

In the sample code, it's mean 78141 bytes of memory was instanciated, but not deleted between heap_trace_start() and heap_trace_stop() );

As sample :
`char *ssid = "testing";`  
means in fact :  
`const char * test = "testing";` String literals are pointers to constant chars

It would be better to use C++ `std::string `instead with auto allocation and deletion. 
The C char[] need to be allocated with `new`, and `delete`

That's a memory leak cause we do not need "ssid" in other place and it was not done to be reused.

**Comment by nliaudat** [comment]
New sample for HEAP_TRACING_CLASS_FLOW_CNN_GENERAL_DO_ALING_AND_CUT

ClassFlowCNNGeneral.cpp:468

```
bool ClassFlowCNNGeneral::doFlow(string time)
{

#ifdef HEAP_TRACING_CLASS_FLOW_CNN_GENERAL_DO_ALING_AND_CUT
    //register a buffer to record the memory trace
    ESP_ERROR_CHECK( heap_trace_init_standalone(trace_record, NUM_RECORDS) );
    // start tracing
    ESP_ERROR_CHECK( heap_trace_start(HEAP_TRACE_LEAKS) );
#endif

    if (disabled)
      return true;

    if (!doAlignAndCut(time)){
        return false;
    };

    LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "doFlow after alignment");

    doNeuralNetwork(time);

    RemoveOldLogs();

#ifdef HEAP_TRACING_CLASS_FLOW_CNN_GENERAL_DO_ALING_AND_CUT
    ESP_ERROR_CHECK( heap_trace_stop() );
    heap_trace_dump(); 
#endif   

    return true;
}
```

**Comment by caco3** [comment]
Thanks  @nliaudat for your adding!

Document in the readme how to use it?
Like we did to show how to build it in https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/README.md#compile

I do not like the way you create the char in your example (`char *ssid = "testing";`).
I never have done it this way and also it can create some undefined behaviour when one wants to modify it, see https://onlinegdb.com/PfxAOwn39, where there is no crash but also no 2nd line printed.

I always would prefer to write `char ssid[] = "testing"` instead.

See also https://stackoverflow.com/questions/10186765/what-is-the-difference-between-char-array-and-char-pointer-in-c

This way `ssid` gets placed on the stack and free whenever the function returns. So there is no memory leak in such case.

In your example I am actually not sure. Is `ssid` stored in the flash (like a `const char)`? If so, then it also will not leak RAM...

**Comment by nliaudat** [comment]
`#define CONFIG_HEAP_TRACING_STANDALONE`

**or better, use the dedicated env :  [env:esp32cam-dev-task-analysis]**

in the file_to_analyse.cpp : 

```
#ifdef CONFIG_HEAP_TRACING_STANDALONE
// or the name of tracing define : #if defined HEAP_TRACING_MAIN_WIFI || defined HEAP_TRACING_MAIN_START
    #include <esp_heap_trace.h>
    #define NUM_RECORDS 300
    static heap_trace_record_t trace_record[NUM_RECORDS]; // This buffer must be in internal RAM
#endif
```
Init the tracing at the start of the analysed function : 

```
    #ifdef CONFIG_HEAP_TRACING_STANDALONE
        //register a buffer to record the memory trace
        ESP_ERROR_CHECK( heap_trace_init_standalone(trace_record, NUM_RECORDS) );
    #endif
```
Start tracing : 

```
    #ifdef CONFIG_HEAP_TRACING_STANDALONE
        ESP_ERROR_CHECK( heap_trace_start(HEAP_TRACE_LEAKS) );
    #endif
```


[Some code to trace]

Stop the trace and dump the records to printf

```
    #ifdef CONFIG_HEAP_TRACING_STANDALONE
        ESP_ERROR_CHECK( heap_trace_stop() );
        heap_trace_dump(); 
    #endif
```

**Comment by nliaudat** [comment]
> I do not like the way you create the char in your example (`char *ssid = "testing";`).

I have invented nothing. It was like that [before](https://github.com/jomjol/AI-on-the-edge-device/commit/17eb382b66ffbff36f987db334c01e2b6fcfd99f#diff-2773902c8e6d0ea3798e839063d7b39aa9b94a2e27c405ae00cadf0ee66ca2de) : 

```
char *ssid = NULL, *passwd = NULL, *hostname = NULL, *ip = NULL, *gateway = NULL, *netmask = NULL, *dns = NULL; int rssithreashold = 0;
    LoadWlanFromFile(WLAN_CONFIG_FILE, ssid, passwd, hostname, ip, gateway, netmask, dns, rssithreashold);

wifi_init_sta(ssid, passwd, hostname, ip, gateway, netmask, dns, rssithreashold);  
```

**Comment by caco3** [comment]
> I have invented nothing. It was like that before

I know. I just wanted to say that I never used that convention to initialize a char array.
Thinking how to interpret `char *a = "abcd";`, I think it means to have an array with "abcd" in flash and create a char pointer `*a` which points to that flash memory. One can then modify the pointer to point somewhere else, but there is no way to modify it.
so it would mean the same as
```
char b[] = "abcd";
char *a = a;
```

But thats my ANSI C knowledge, maybe in C++ this works differently.

PR #1877, opened 2023-01-19
Show WIFI signal text labels / Log RSSI value to logfile

Slider0007:
1. Show WIFI signal text label for different ranges in addition to pure RSSI value to make it more transparent what's a good signal quality.

Proposed labels:
- WIFI Signal: Excellent (RSSI value)
- WIFI Signal: Good (RSSI value)
- WIFI Signal: Fair (RSSI value)
- WIFI Signal: Weak (RSSI value)
- WIFI Signal: Unreliable (RSSI value)

image

2. Log RSSI value to logfile at the end of each round (diagnostic purpose, debug loglevel)

### DIFF
--- a/code/components/jomjol_tfliteclass/server_tflite.cpp
+++ b/code/components/jomjol_tfliteclass/server_tflite.cpp
@@ -870,12 +870,12 @@ void task_autodoFlow(void *pvParameter)
             LogFile.RemoveOldDataLog();
         }
         
-        //CPU Temp
-        float cputmp = temperatureRead();
+        //CPU Temp -> Logfile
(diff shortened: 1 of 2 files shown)


caco3:
Instead of text labels, why not use an icon like known from the smartphone or laptop?

Slider0007:
@caco3: Thanks for your feedback. I was also thinking about icons but the actual design is more and less a pure textual presentation so I thought a it would fit better to have it also text based. If using icons I think the page should be rearanged somehow... We can keep this in mind for future improvement, though -> e.g. having a kind of status line which is always present on every page...

Issue #1887, opened 2023-01-21
REST-API: /sysinfo not in json ?


ozett:
### The Problem

try too make some REST-api calls on some endpoints,
but got errors for the format.

is there something wrong with the syntax of the expected json-return?
https://github.com/jomjol/AI-on-the-edge-device/blob/defbd60ccf8ca8791df0a14c392e6f0ac65ee5a1/code/main/server_main.cpp#L340

### Version
Release: v13.0.8 (Commit: 04f2f23+)
### Logfile
```shell
-
```
### Expected Behavior
no errors on json-syntax

### Screenshots
| 1 |
| :---:|
|![image](https://user-images.githubusercontent.com/5937811/213864182-0abb004e-fa4a-4ece-b3b4-018f0316e593.png)|
| 2 |
![image](https://user-images.githubusercontent.com/5937811/213864262-3ced74d3-53a2-42b5-952a-1a7e55c8d226.png)
|3|
![image](https://user-images.githubusercontent.com/5937811/213866410-d5b5e4d7-4135-4414-88a5-9df1b3f399c5.png)

### Additional Context

[{"id":"78c1700e5af5bb63","type":"inject","z":"d838f7c4a438035b","name":"rest-api","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":1630,"y":420,"wires":[["1d4498daebf8ce3a"]]},{"id":"1d4498daebf8ce3a","type":"http request","z":"d838f7c4a438035b","name":"63-sysinfo","method":"GET","ret":"obj","paytoqs":"ignore","url":"http://192.168.14.63/sysinfo","tls":"","persist":false,"proxy":"","insecureHTTPParser":true,"authType":"","senderr":false,"headers":[],"x":1830,"y":420,"wires":[["ceecbd5a353a2499"]]},{"id":"ceecbd5a353a2499","type":"debug","z":"d838f7c4a438035b","name":"sysinfo","active":true,"tosidebar":true,"console":false,"tostatus":true,"complete":"true","targetType":"full","statusVal":"payload","statusType":"auto","x":2180,"y":420,"wires":[]}]

ozett:
this is great return on missing endpoints.
sadyl seems not to work on /info

[...]

PR #1908, opened 2023-01-23
REST handler CPU temp: escape special character

Slider0007:
REST handler CPU temp: Escape special character

image

Issue reported here https://github.com/jomjol/AI-on-the-edge-device/issues/1887#issuecomment-1399232082

### DIFF
--- a/code/components/jomjol_tfliteclass/server_tflite.cpp
+++ b/code/components/jomjol_tfliteclass/server_tflite.cpp
@@ -692,7 +692,7 @@ esp_err_t handler_cputemp(httpd_req_t *req)
     const char* resp_str;
     char cputemp[20];
     
-    sprintf(cputemp, "%4.1f°C", temperatureRead());
+    sprintf(cputemp, "%4.1f&degC", temperatureRead());
 
     resp_str = cputemp;
 


caco3:
I am not sure if we should escape it, and how would you do it?

All it is is an **extended** ASCII character (-> https://theasciicode.com.ar/extended-ascii-code/degree-symbol-ascii-code-248.html) and the browser does not render it correctly.
In the Web UI and also in Bash and in the webbrowser Dev Tools it is shown correctly!

Slider0007:
@caco3: There some options to potentially handle this:
- keep it as default HTML output, than I think we should use HTML character &deg for °
- switch to plain/text and use extended ASCIIs (charset=utf-8 needs to be included in header)
- remove units in general and provide just numbers (would be my prefered solution)

caco3:
> remove units in general and provide just numbers (would be my prefered solution + not having these issues at all)

I agree to this variant.
And add the `°C` directly to the WebUI.

Note: As far as I can see, the temperature is always in °C`. I am just waiting for the first American who wants it in °F` :)

Slider0007:
Thanks for your reply. I'm going to prepare the handler with just pure number output and unit be added in UI.
I'm going to to prepare the same for RSSI handler where also the unit is sent with right now.

> ` I am just waiting for the first American who wants it in °F` :)

That could be the case at some time ;-)

Slider0007:
@caco3: Updated implementation and PR description with changes prepared

caco3 (approved):

PR #1924, opened 2023-01-27
Option alignment algo off

nliaudat:
Add the option to disable alignment algo.  Tested on real board without problem.

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp
@@ -65,7 +65,7 @@ bool ClassFlowAlignment::ReadParameter(FILE* pfile, string& aktparamgraph)
     std::vector<string> splitted;
     int suchex = 40;
     int suchey = 40;
-    int alg_algo = 0;
+    int alg_algo = 0; //default=0; 1 =HIGHACCURACY; 2= FAST; 3= OFF //add disable aligment algo |01.2023
 
 
     aktparamgraph = trim(aktparamgraph);
@@ -130,6 +130,8 @@ bool ClassFlowAlignment::ReadParameter(FILE* pfile, string& aktparamgraph)
                 alg_algo = 1;
             if (toUpper(splitted[1]) == "FAST")
                 alg_algo = 2;
+            if (toUpper(splitted[1]) == "OFF") //no align algo if set to 3 = off => no draw ref //add disable aligment algo |01.2023
+                alg_algo = 3;
         }
     }
 
@@ -145,7 +147,10 @@ bool ClassFlowAlignment::ReadParameter(FILE* pfile, string& aktparamgraph)
         #endif
     }
 
-    LoadReferenceAlignmentValues();
+    //no align algo if set to 3 = off => no draw ref //add disable aligment algo |01.2023
+    if(References[0].alignment_algo != 3){
+        LoadReferenceAlignmentValues();
+    }
     
     return true;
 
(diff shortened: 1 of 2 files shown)


jomjol:
Let us wait until the v14.0.0 is finally released.

nliaudat:
After 545 rounds with no alignment algo, no problem to signal. 
It works fine without alignment algo. 
The analog model was not tested as I have no watermeter with analog.

caco3:
set to draft to prevent accidental merge before we released the final v14

nliaudat:
> set to draft to prevent accidental merge before we released the final v14

Why not release in 14 ? 
The user has to change the config in expert mode to set alignment off. 
No risk to break upgrade with 13.8

caco3:
What we have now in rolling got tested by many people.

> No risk to break upgrade with 13.8

There is always a risk.

Nothing stops us from creating a 14.1 in the near future!

nliaudat:
The alignment algo OFF function is a proof of concept to show you that doing post alignment is a bad thing. 

It needs the whole image to align and takes a lot of memory. The after alignment result is sometimes worse than the original. 

The best way to go is : 

take image > rotate in raw format* > cut in the whole ROI regions or subdivided in ROIs per value > save as jpeg

* rotating a jpeg image cause many blur and distortion cause it's a compressed format. There is no way to rotate jpeg images without loose with embedded devices like esp32.

nliaudat:
téléchargement (2)

After alignment alignment algo fast : 
alg

jomjol:
This is not a real problem as the images get downsized quite a lot and the CNN should handle this artifacts.

nliaudat:
You've made 7 RC 14.0 release since that PR .... And you do not want to test it since release 14.1 .

You don't want admit the alignment algo is useless ?

friedpa:
Sorry @nliaudat what do you want to tell us? if you want something to be implemented simply WAIT, or implement it yourself in your Fork. For me the alignment algo works fine.
The developers (in their spare time) are working with a roadmap, if you like it or not.

caco3:
> You've made 7 RC 14.0 release since that PR .... And you do not want to test it since release 14.1 .

Did you understand why we did 8 RCs?
We had an issue with the creation of the zip files. They did not fully have the content we wanted due to caching. also we removed some no longer needed files in them.
And the way the pipeline currently is, it only could be tested with a release.

caco3:
I am removing the draft state again (since I have set it), but I let it open to @jomjol if/when to merge it.

caco3:
I am removing the draft state again (since I have set it), but I let it open to @jomjol if/when to merge it.

jurasjo:
> The developers (in their spare time) are working with a roadmap, if you like it or not.

Do you realize that @nliaudat is also quite active contributor to this project? Right?
https://github.com/jomjol/AI-on-the-edge-device/graphs/contributors

jurasjo:
> Sorry @nliaudat what do you want to tell us? if you want something to be implemented simply WAIT, or implement it yourself in your Fork. For me the alignment algo works fine. The developers (in their spare time) are working with a roadmap, if you like it or not.

Do you realize that @nliaudat is also quite active contributor to this project? Right?
https://github.com/jomjol/AI-on-the-edge-device/graphs/contributors

jomjol:
I will merge it to rolling immediatly as we are now in the next round of features. But just one remark, why the alignment algo is usefull from my point of view and why I will not recommend to turn it of:

I can remove and add my device at any time and it still will work without any retraining, as the minor changes in position and rotation are corrected by the alignment algo. With this algo I still can read out every two minutes, which is more than often enough.

At least in my home I have to remove the device minimum once a year, when somebody offical comes to check read out my meters. As a developer even more often :-)

nliaudat:
Many thanks for merging in rolling, but the option is not available in release 14.0.3.
Are the release are build upon master branch ?

I understand your needs for the alignment algo, but should it not be off or fast by default ?
Regards

caco3:
Like we already explained, we did not want to have it in the v14 release.
it is now in rolling and will end up in the next version (whatever it will be).

jomjol:
> I understand your needs for the alignment algo, but should it not be off or fast by default ?
> Regards

No, because first of all we want a stable device. The main benefit would be a speed up, but this is not really a major need.

jurasjo:
I use this feature. Is it included in the final v15.1.0 or do I have to use rolling?

caco3:
all of current rolling is in 15.1

jurasjo:
Thank you.

PR #1967, opened 2023-02-02
Cleanup defines.h

Slider0007:
- Disable define USE_HIMEM_IF_AVAILABLE because function generally is disabled in sdkconfig.defaults
- Remove orphan define ALGROI_LOAD_FROM_MEM_AS_JPG__SHOW_TAKE_IMAGE_PROCESS (not used anymore)

### DIFF
--- a/code/include/defines.h
+++ b/code/include/defines.h
@@ -21,7 +21,9 @@
 
 
     // use himem //https://github.com/jomjol/AI-on-the-edge-device/issues/1842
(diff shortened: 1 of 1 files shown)


nliaudat:
This would be better : 

```
    #if (CONFIG_SPIRAM_BANKSWITCH_ENABLE)
            #define USE_HIMEM_IF_AVAILABLE 1
    #endif
```

Slider0007:
@nliaudat: I added your suggestion to the PR

PR #1986, opened 2023-02-05
Generate all parameter documentation from the documentation repo

caco3:
- Replaced the parameter information with tooltips
- The tooltips get generated at build time as separate html files
- The source of the tooltips is the online documentation at https://jomjol.github.io/AI-on-the-edge-device-docs/Parameters/

grafik

# TODO
- [ ] rebase after https://github.com/jomjol/AI-on-the-edge-device/pull/2001 got merged
- [ ] update parameters and parameter docs based on https://github.com/jomjol/AI-on-the-edge-device/pull/2001
- [ ] Use nicer names for the first column in the config page (`LogImageLocation` -> `Log Image Location`).


jomjol:
I'm not sure, if we should remove this completly from the html-page. As a first quick overview I experienced them very usefull, e.g. in which direction is the sign of the decimal shift, or how should an adress look like, ...
I would suggest to keep a minimum docu there and implement the link to the extended description in the docu, which is an excelent description maintained by @caco3 

Other opinions?

caco3:
I would like to have to maintain a single place for the documentation. Else it will be very hard to keep it up-to-date.
I was thinking of embedding the documentation of each parameter in a tooltip, like eg. shown in http://www.menucool.com/tooltip/tooltip/demo2.html. But I did not get it working yet.

caco3:
What do you think about this:
grafik

When you hover over the blue button, a tooltip gets shown.
I have to work on the styling (size, scrollbars, ...), but in general it works.

The tooltips get generated from the same source as the documentation.

jomjol:
This would be a good compromise and only exclude the people, who do not give access to the internet. But that is okay for me.

Slider0007:
Cool, this looks really promising... I'm exited to see more of it!

caco3:
Please let me know what you think

https://github.com/jomjol/AI-on-the-edge-device/actions/workflows/build.yaml?query=branch%3Amove-parameter-doc-to-online

Each tooltip has its own html file. Currently they are all loaded when the page gets loaded (around 100 files(!)).
I am not sure if this is the right way to do it, but it is the easiest and the files will get cached in the webbrowser.
Alternatively I would have to use javascript to trigger the html file download of a tooltip whenever you do a mouseover or touche the help icon. Since this takes time and can also fail, the tooltip would be empty.

caco3:
Feel free to play with this PR, it works as supposed.
How ever before we merge it, I want to merge https://github.com/jomjol/AI-on-the-edge-device/pull/2001 and then rebase this.

jomjol:
Sorry, guy, lost the overview how to implement new parameters. Currently I cannot clone the doc repos (see issue there (https://github.com/jomjol/AI-on-the-edge-device-docs/issues/17). 
I have added a new interface to InfluxDB v2 - how do I update the parameter documentation?

caco3:
> I have added a new interface to InfluxDB v2 - how do I update the parameter documentation?

See https://github.com/jomjol/AI-on-the-edge-device-docs/issues/17#issuecomment-1424795047 to update the documentation.

caco3:
replaced by https://github.com/jomjol/AI-on-the-edge-device/pull/2059

PR #2001, opened 2023-02-08
*** OLD *** Migrate configuration -> Superseeded by PR2017

caco3:
Replaced by https://github.com/jomjol/AI-on-the-edge-device/pull/2017

---

This PR adds a migration function of some parameters:
- Rename section `MakeImage` to `TakeImage`, rename all variables and files accordingly
- Rename `Intervall` to `Interval`, rename all variables
- Rename `RSSIThreashold` to `RSSIThreshold`, rename all variables
- For all boolean parameters: If they where commented out -> remove the semicolon and set them to their default value (according to their C++ initialization).

The migration function gets called at startup and modifies the config file as needed. If one or mor eparameter needs a migration, store the `config.ini` as `config.bak` before any change to it.

# Yet to do
- [x] Remove the checkboxes for all boolean parameters in the UI -> @jomjol can you do this? I tried it but failed to understand how it exactly works.
- [ ] @jomjol @haverland @Slider0007: Are there other parameters where you think we should rename them? This PR might be a good time to do so.
- [ ] Testing

### DIFF
--- a/FeatureRequest.md
+++ b/FeatureRequest.md
@@ -180,9 +180,9 @@ haveing this state in the mqtt broker can trigger functions like closing the ate
 
 * Implementation of an authentication mechanism.
 
-#### #8 MQTT configurable readout intervall
+#### #8 MQTT configurable readout interval
 
-Make the readout intervall configurable via MQTT.
+Make the readout interval configurable via MQTT.
 
 * Change the mqtt part to receive and process input and not only sending
 
--- a/code/components/jomjol_fileserver_ota/server_file.cpp
+++ b/code/components/jomjol_fileserver_ota/server_file.cpp
@@ -996,8 +996,13 @@ std::string unzip_new(std::string _in_zip_file, std::string _target_zip, std::st
 
                 ESP_LOGI(TAG, "Filename to extract: %s, Zwischenfilename: %s", zw.c_str(), filename_zw.c_str());
 
+                std::string folder = path.substr(0, filename_zw.find_last_of('/'));
+                MakeDir(folder);
+
                 // extrahieren in zwischendatei
                 DeleteFile(filename_zw);
+
+
                 FILE* fpTargetFile = fopen(filename_zw.c_str(), "wb");
                 uint writtenbytes = fwrite(p, 1, (uint)uncomp_size, fpTargetFile);
                 fclose(fpTargetFile);
--- a/code/components/jomjol_fileserver_ota/server_file.cpp
+++ b/code/components/jomjol_fileserver_ota/server_file.cpp
@@ -739,10 +739,6 @@ static esp_err_t upload_post_handler(httpd_req_t *req)
     /* Redirect onto root to see the updated file list */
     httpd_resp_set_status(req, "303 See Other");
     httpd_resp_set_hdr(req, "Location", directory.c_str());
-
-    /* Redirect onto root to see the updated file list */
-    httpd_resp_set_status(req, "303 See Other");
-    httpd_resp_set_hdr(req, "Location", directory.c_str());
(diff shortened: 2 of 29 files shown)


jomjol:
> @jomjol @haverland @Slider0007: Are there other parameters where you think we should rename them? This PR might be a good time to do so.

I started to do so, but as the categories are renamed as well, I could not fully test. Will do this in the next days.

caco3:
@jomjol You should start with the `sd-card/html/edit_config_param.html` from this branch, I updated the section/parameters already

caco3:
@jomjol There was a merge conflict, so I solved it, see https://github.com/jomjol/AI-on-the-edge-device/pull/2001/commits/d1d2d67880be2ddb347d8c18674e8e5dfd9680cd

Not sure if I now reverted some of your work!

jomjol:
> @jomjol There was a merge conflict, so I solved it, see [d1d2d67](https://github.com/jomjol/AI-on-the-edge-device/commit/d1d2d67880be2ddb347d8c18674e8e5dfd9680cd)
> 
> Not sure if I now reverted some of your work!

I think I just corrected it.

Slider0007:
@caco3: Some discussion topics:

config.ini:
`AutoAdjustSummertime` --> AFAIK this not used anymore
`LogfileRetentionInDays` --> maybe `LogRetentionPeriod` (specially for logging the images, Logfile... is kind of misleading, but renaming this is not a must)
`SetRetainFlag` --> `RetainFlag` or `RetainMessages`(renaming not a must)

caco3:
> config.ini: `AutoAdjustSummertime` 

Stimmt, das wird in der Firmware nirgends mehr verwendet. @jomjol, können wir das löschen?

> LogfileRetentionInDays --> maybe LogRetentionPeriod (specially for logging the images, Logfile... is kind of misleading, but renaming this is not a must)

Right.
I would suggest to do it as following:
- Analog/Digits: Here we use it for the training images. I would rename it to `RoiImagesRetentionInDays` or `TrainingImagesRetentionInDays`. And I would rename the related parameters as well. Ans also update the default path (eg. `/log/analog`)
- Debug: Keep `LogfileRetentionInDays` as is. 

> `SetRetainFlag` --> `RetainFlag` or `RetainMessages`(renaming not a must)

I would go for `RetainMessages`

@jomjol, what is `LogImageLocation` in the section `TakeImage` used for? Is that storning the whole cam image?

jomjol:
> > config.ini: `AutoAdjustSummertime`
> 
> Stimmt, das wird in der Firmware nirgends mehr verwendet. @jomjol, können wir das löschen?

Right, we can delete it


> > LogfileRetentionInDays --> maybe LogRetentionPeriod (specially for logging the images, Logfile... is kind of misleading, but renaming this is not a must)
> 
> Right. I would suggest to do it as following:
> 
>     * Analog/Digits: Here we use it for the training images. I would rename it to `RoiImagesRetentionInDays` or `TrainingImagesRetentionInDays`. And I would rename the related parameters as well. Ans also update the default path (eg. `/log/analog`)
TrainingImagesRetention... could make the people think, the device is training itself. Therefore I would go for "ROIImages..."


> @jomjol, what is `LogImageLocation` in the section `TakeImage` used for? Is that storning the whole cam image?

This is used for storing the full images (before cutting) in the directory `/log/source`

caco3:
I think we should to a release 14.1 which just the parameter migration.
Release 15 will have a lot of changes which might break stuff. And this way users can go back top v14 without broken parameters.

I will prepare a branch for it.

caco3:
New PR which **only** contains the parameter changes: https://github.com/jomjol/AI-on-the-edge-device/pull/2017/

Issue #1851, opened 2023-01-15
Support saturation / contrast or at least check actual support to have proper documentation


Slider0007:
### The Feature

Support saturation / contrast or at least check actual support to have proper documentation

![image](https://user-images.githubusercontent.com/115730895/212561781-94ef120a-d763-47c8-83f8-cfd460c9a2e5.png)

comapre discussion #1826

jomjol:
This is a problem of the copied ESP32 driver, which we will not update within this project. We can update it to the most recent one.
But as there is a high risk (out of experience), that we run into stability problems. Therefore we should do that only after the next release.

Slider0007:
Implemented with #2048.

Please be aware that this change could affect existing camera configuration.

PR #2048, opened 2023-02-18
Support saturation & contrast

Slider0007:
Add saturation & contrast support --> #1851

- Add workaround for cam library to support saturation and contrast
--> Not all requested functions are enabled even if called because bitwise OR logic of the ENABLE BYTE is missing in library. Only latest enable is active. Now registers are written directly to enable neccessary bits. Newer versions of cam library still have the same issue, so same library version is used
- Add additional error handling to ClassControllCamera
- Refactor ClassControllCamera.cpp
- edit_reference.html: Use correct value for contrast setting (saturation value was used before)
- Limit image quality index to 8...63 to prevent against instabilities at too low values (no change to default value: 12)
- edit_config_param.html: Adapt parameter quality index, saturation and contrast to changes in firmware

Questions:
- [x] Any concerns about this workaround?
- [x] Any concerns changing the lower limit of quality index from 0 to 8?

### DIFF
--- a/code/components/jomjol_controlcamera/ClassControllCamera.cpp
+++ b/code/components/jomjol_controlcamera/ClassControllCamera.cpp
@@ -63,7 +63,6 @@ static camera_config_t camera_config = {
     .fb_count = 1,       //if more than one, i2s runs in continuous mode. Use only with JPEG
     .fb_location = CAMERA_FB_IN_PSRAM, /*!< The location where the frame buffer will be allocated */
     .grab_mode = CAMERA_GRAB_LATEST,      // only from new esp32cam version
-    
 };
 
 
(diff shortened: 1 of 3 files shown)


jomjol:
To your questions
1) no concerns - only one question: do we have to take something into account, when we update the esp32cam library?
2) fine for me!

For me you can remove the draft status.

Slider0007:
> 1. no concerns - only one question: do we have to take something into account, when we update the esp32cam library?

It should be not critical to update the library because the same registers get written than the library do.

Slider0007:
@jomjol: We should highlight this change somehow on the changelog because this could have impact on existing camera configurations.

jomjol:
Done in the changelog from the rolling

PR #2070, opened 2023-02-23
Restructure webui menu

caco3:
I reordered the menu a bit.
IMO it is now more logic.

@jomjol @Slider0007 What do you think?

I was even think if we should move the `Settings > Set previous Value` to `Data > Set previous Value`. But it would be the only menu item there where one actively would set something.

grafik

grafik

### DIFF
--- a/sd-card/html/index.html
+++ b/sd-card/html/index.html
@@ -82,15 +82,22 @@
         </li>
       </ul>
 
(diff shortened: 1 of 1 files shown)


Slider0007:
> @jomjol @Slider0007 What do you think?
> 
> I was even think if we should move the `Settings > Set previous Value` to `Data > Set previous Value`. But it would be the only menu item there where one actively would set something.

I like your proposal and would keep "set previous value" in settings menu

PR #2079, opened 2023-02-25
Migrate some UART logs to Logfile logs

caco3:
Migrate some UART logs to Logfile logs for easier troubleshooting in case of an error.

Note: As pointed out below, the initial log level is INFO. This means we can not migrate the DEBUG logs as the will not be shown in the logfile and neither on UART.

- Migrated all ERROR logs
- migrated some INFO logs


caco3:
@jomjol This would have made the dection of the crash in https://github.com/jomjol/AI-on-the-edge-device/pull/2080 easier :)

I tried it also for the other log levels (see https://github.com/jomjol/AI-on-the-edge-device/pull/2078), but the it hangs somewhere.

Slider0007:
> I tried it also for the other log levels (see #2078), but the it hangs somewhere.

In early boot phase (also during OTA update) default log level is set to INFO. This means DEBUG level won't get visible until we do not change default log level. I also played with this but setting to DEBUG would somehow mess and spam the boot log...

caco3:
Thanks!

My problem was that the USB log no longer worked. The default level of INFO explains it as it then also get suppressed on USB.

PR #2080, opened 2023-02-25
Keep the html folder on a flat structure (no subfolders)

caco3:
Else the OTA fails until >15.0.x because it does not create the needed subfolders:

```
I (130121) OTA FILE: Filename to extract: /sdcard/html/param_tooltips/flipImageSize.png, Zwischenfilename: /sdcard/html/param_tooltips/flipImageSize.png_0xge
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.
```

Note: In current roling this is already fixed.


caco3:
@jomjol @Slider0007 

If you updated from 15.0.1 or older to the recent rolling, the firmware update would fail if the folder `/sdcard/html/param_tooltips` was missing.

PR #2083, opened 2023-02-25
do not show value on recognition page

caco3:
@jomjol On some of my devices I see on top of the recognition page a value (probably the raw value):
grafik

I have no clue why it is there and why it is only on some of my devices.

This PR tries to fix this, can you check if it is ok?


jomjol:
I checked it with my water meter and it looks good:
grafik

Let's merge it.

P.S.: I did not get fully, what you meant with the number on top.

caco3:
> P.S.: I did not get fully, what you meant with the number on top.

I did not want to have the value to be shown at the top of the page.
But I prepended a header now so its ok :)

Issue #1886, opened 2023-01-21
Implement some SD card basic check in early device init phase


Slider0007:
### The Feature

Implement some SD card basic check in early device init phase

Some ideas:
- Write to SD card and read back
- Check basic folder structure
- User feedback if some condition is not met e.g. blinking error codes

Slider0007:
Implemented with #2085 into rolling branch. This will be part of next official release 15.1

PR #2085, opened 2023-02-26
SD card basic R/W check + folder/file presence check

Slider0007:
Implement a basic SD card check at boot to ensure proper SD card functionaliy -> feature request #1886 

- check function 1:
  - Create file
  - Write file with some generic text
  - Read file back
  - Verify CRC
  - Delete file
  --> After error occured it boots to reduced WebUI show and error code

- check function 2:
  - Verify presence of some folders / files
  - The following folders / files get check at boot:
    /sdcard/config
    /sdcard/html
    /sdcard/demo --> created automatically in firmware
    /sdcard/firmware --> created automatically in firmware
    /sdcard/img_tmp --> created automatically in firmware
    /sdcard/log --> created automatically in firmware
    /sdcard/wlan.ini
    /sdcard/config/config.ini
    /sdcard/html/index.html
    /sdcard/html/ota_page.html
    /sdcard/html/log.html
    /sdcard/html/common.js
    /sdcard/html/gethost.js
    /sdcard/html/version.txt
     --> If one or more are not present -> load reduced WebUI and show error code

- Check indication in boot log (all OK):
image

- LED indication for the respective error cases:
image

###################################################
Questions:

- [X] Shall we stop the boot process in this early phase when SD card RW check fails or is it better to continue to reduced web interface?
  --> Load reduced web interface and show error code
- [X] Presence check of some folders is not garantuee that everthing on SD is OK, but could give early indication that something is at least present (only warnings, no reaction). What do you think about this? Does this make sense?
  --> Load reduced web interface and show error code

TODO:
- Update documentation

### DIFF
--- a/code/components/jomjol_helper/Helper.h
+++ b/code/components/jomjol_helper/Helper.h
@@ -76,6 +76,8 @@ enum SystemStatusFlag_t {          // One bit per error
     SYSTEM_STATUS_PSRAM_BAD         = 1 << 0, //  1, Critical Error
     SYSTEM_STATUS_HEAP_TOO_SMALL    = 1 << 1, //  2, Critical Error
     SYSTEM_STATUS_CAM_BAD           = 1 << 2, //  4, Critical Error
+    SYSTEM_STATUS_SDCARD_CHECK_BAD  = 1 << 3, //  8, Critical Error
+    SYSTEM_STATUS_FOLDER_CHECK_BAD  = 1 << 4, //  16, Critical Error
 
     // Second Byte
     SYSTEM_STATUS_CAM_FB_BAD        = 1 << (0+8), //  8, Flow still might work
--- a/code/components/jomjol_helper/sdcard_check.cpp
+++ b/code/components/jomjol_helper/sdcard_check.cpp
@@ -1,0 +1,112 @@
+#include "sdcard_check.h"
+#include <string.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <sys/stat.h>
+
+#include "esp_rom_crc.h" 
+#include "ClassLogFile.h"
+
+static const char *TAG = "SDCARD";
+
+int SDCardCheckRW(void)
+{
+    LogFile.WriteToFile(ESP_LOG_INFO, TAG, "Basic R/W check started...");
+    FILE* pFile = NULL;
+    int iCRCMessage = 0;
+   
+    pFile = fopen("/sdcard/sdcheck.txt","w");
+    if (pFile == NULL) {
+        LogFile.WriteToFile(ESP_LOG_ERROR, TAG, "Basic R/W check: (E1) No able to open file to write");
+        return -1;
+    } 
+    else {
+        std::string sMessage = "This message is used for a SD-Card basic check!";
+        iCRCMessage = esp_rom_crc16_le(0, (uint8_t*)sMessage.c_str(), sMessage.length());
+        if (fwrite(sMessage.c_str(), sMessage.length(), 1, pFile) == 0 ) {
+            LogFile.WriteToFile(ESP_LOG_ERROR, TAG, "Basic R/W check: (E2) Not able to write file");
+            fclose(pFile);
+            unlink("/sdcard/sdcheck.txt");
+            return -2;
+        }
+        fclose(pFile); 
+    }
+
+    pFile = fopen("/sdcard/sdcheck.txt","r");
+    if (pFile == NULL) {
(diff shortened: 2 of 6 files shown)


jomjol:
> Shall we stop the boot process in this early phase when SD card RW check fails or is it better to continue?

Yes, we should do this, as this is a basic need of the firmware.

> Presence check of some folders is not garantuee that everthing on SD is OK, but could give early indication that something is at least present (only warnings, no reaction). What do you think about this? Does this make sense?

It does make sense and we can write a problem into the log file. I would  not stop the firmware in this case

Slider0007:
> > Shall we stop the boot process in this early phase when SD card RW check fails or is it better to continue?
> 
> Yes, we should do this, as this is a basic need of the firmware.
> 
> > Presence check of some folders is not garantuee that everthing on SD is OK, but could give early indication that something is at least present (only warnings, no reaction). What do you think about this? Does this make sense?
> 
> It does make sense and we can write a problem into the log file. I would not stop the firmware in this case

Thanks for your feedback. Implementation completed:
- Failed SD card basic check --> abort booting
- Folder check --> continue booting, only warning

caco3:
>      Shall we stop the boot process in this early phase when SD card RW check fails or is it better to continue?
> Yes, we should do this, as this is a basic need of the firmware.

We could instead show a basic website (contained in the firmware) which documents the issue.

And like an idea I had before, we also could contain a basic OTA page so one could still update in case the web UI got deleted.

caco3:
I actually already implemented this for the index page: https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/main/server_main.cpp#L236

caco3:
keep in mind, if the web UI does not initialized, the users can only check the log using USB!

Slider0007:
> > ```
> >  Shall we stop the boot process in this early phase when SD card RW check fails or is it better to continue?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Yes, we should do this, as this is a basic need of the firmware.
> 
> 
> We could instead show a basic website (contained in the firmware) which documents the issue.

@caco3: Thanks for reminding to the reduced web interface. I wasn't thinking about this during design, but I like the idea to have the chance to get remote visual feedback in error case. If we do not get to this point, we still have at least local indication (console log and board status LED).

@caco3, @jomjol: The updated design approach could be to following:

- SD card basic check failed (not able to read, write or delete a file on SD card):
  --> Continue booting and open reduced web interface
  --> Recommendation in the error description: Check / format or replace SD card (because with such an error case it's potentially impossible to recover with only doing OTA)
  --> Does it makes sense to split the error to write, read and delete errors or combine them and just having one error?

- SD basic folder structure check (check availability of some folders / files):
  --> Continue booting and also open only reduced web interface
  --> Recommendation in the error description: Check log and redo OTA

caco3:
> * SD card basic check failed (not able to read, write or delete a file on SD card):
>   --> Continue booting and open reduced web interface
>   --> Recommendation in the error description: Check / format or replace SD card (because with such an error case it's potentially impossible to recover with only doing OTA)
>   --> Does it makes sense to split the error to write, read and delete errors or combine them and just having one error?

I think we could keep the reduced web page minimalistic (to save flash memory).
It should be enough to show the error code and a link to https://jomjol.github.io/AI-on-the-edge-device-docs/Error-Codes/

With just working read we **could** continue with the normal UI, but I **would not** as it could be misleading.

> * SD basic folder structure check (check availability of some folders / files):
>   --> Continue booting and also open only reduced web interface
>   --> Recommendation in the error description: Check log and redo OTA

I would also just go to a reduced web page.
If you want I can write a minimalistic OTA page which could be contained as a fallback.

Files which IMO should be available (resp. checked for):
- config.ini -> although the device should work without it
- wlan.ini
- index.html
- ota.html
- Log.html
- common.js
- gethost.js
- version.txt (not really needed, but a good indication that the Web UI got properly installed. Some users just copy it from the repo but that does no longer work as we modify the Web UI during the build)

and should we check for at least one tfl file each (analog/digit)?

Slider0007:
@caco3: Thanks for your input. I revised the logic to benefit from reduced web interface.

@caco3, @jomjol: 
- After SD R/W error occured it boots to reduced WebUI show and error code.
- The following folders / files get check at boot. If one or more are not present -> load reduced WebUI and show error code:

/sdcard/config
/sdcard/html
/sdcard/firmware --> created automatically in firmware
/sdcard/img_tmp --> created automatically in firmware
/sdcard/log --> created automatically in firmware
/sdcard/demo --> created automatically in firmware
/sdcard/wlan.ini
/sdcard/config/config.ini
/sdcard/html/index.html
/sdcard/html/ota_page.html
/sdcard/html/log.html
/sdcard/html/common.js
/sdcard/html/gethost.js
/sdcard/html/version.txt

PR #2091, opened 2023-02-27
Improve MQTT

caco3:
- Check that all topics sent successfully
- Only send Homeassistant Discovery the first time we connect. If some topics fail, it retries in the next round. Once all topics got set successfully, it no longer sends them automatically (still possible through the REST API).
- Fixed default values if no MQTT Client ID or maintopic got defined in config.
- Default MQTT Client ID is now the hostname (was "AIOTED-<MAC>", but that is longer tan the allowed 23 bytes)

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
@@ -255,7 +255,7 @@ bool ClassFlowMQTT::doFlow(string zwtime)
                 success |= MQTTPublish(namenumber + "error", resulterror, SetRetainFlag);
 
             if (resultrate.length() > 0) {
-                MQTTPublish(namenumber + "rate", resultrate, SetRetainFlag);
+                success |= MQTTPublish(namenumber + "rate", resultrate, SetRetainFlag);
                 
                 std::string resultRatePerTimeUnit;
                 if (getTimeUnit() == "h") { // Need conversion to be per hour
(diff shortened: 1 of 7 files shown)


Slider0007:
> * Only send Homeassistant Discovery the first time we connect. If some topics fail, it retries in the next round. Once all topics got set successfully, it no longer sends them automatically (still possible through the REST API).

Good to see that you're addressing the MQTT HA discovery topic :-)
I just briefly checked your approach. This is for sure a step in the right direction and improve the situation but unfortunately in worst case it's still remained the same. Reconnect of wifi can happen at any time and if some messages weren't processed at first attempt (which also could happen during flow when wifi is not connection directly after startup) they possibly need to be reprocessed during running flow (GotConnected function gets called after reconnect of wifi and not at the end of each round).
Additionally the REST API can also trigger discovery sending at any time.

To ensure stability at any situation we should take care of worst case scenario.

Having the goal in mind to implement to wifi roaming and have still room for further improvemnts, we should try to avoid sending HA discovery during main flow steps at any kind, so my suggestion is to trigger discovery only when flow is finished with a button. This would be perfectly deterministic and controlable and would garantuee more free heap. IMHO the drawback is really little because discovery is only used once for initial HA setup.

My first heap analysis showed, only ca. 3kB of internal heap left with activated wifi roaming even with tightening some other tasks. This means with activating roaming further implementation are hardly possible without the risk of having same situation than with v13.x. Keeping this HA discovery design like it is even with this optimization I do not recommend wifi roaming implementation due to lack of heap.



> * Fixed default values if no MQTT Client ID or maintopic got defined in config.
> * Default MQTT Client ID is now the hostname (was "AIOTED-", but that is longer tan the allowed 23 bytes)

Cool 👍

caco3:
you are right, I completely missed that point, sry!

My focus was on the missing default values and then the other change was just short but missed the overall picture.

Lets discuss that topic in https://github.com/jomjol/AI-on-the-edge-device/issues/2092

PR #2106, opened 2023-03-01
Logfile: Print start indication block after time sync

Slider0007:
- Print start indication block after time sync occured to highlight start of new logs after e.g. a cold boot (no vaild time available, so all logs before time is synced are printed to 1970-01-01.txt)
- This block is only printed when no time was set before
- The file log_1970-01-01.txt is not getting deleted if NTP time was not set at boot (e.g. after cold boot) because some of the boot logs are in there. If the time is already set during boot, the file will be handled like all other log files and will be deleted after flow is finished.

This topic was raised by @caco3

### DIFF
--- a/code/components/jomjol_logfile/ClassLogFile.cpp
+++ b/code/components/jomjol_logfile/ClassLogFile.cpp
@@ -15,6 +15,7 @@ extern "C" {
 #endif
 
 #include "Helper.h"
+#include "time_sntp.h"
 #include "../../include/defines.h"
 
 static const char *TAG = "LOGFILE";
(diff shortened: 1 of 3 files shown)


caco3:
That looks good to me!

And I now realize, why I never saw the 1970-0101.txt logfile:
It simply gets deleted as it is far to old, see https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_logfile/ClassLogFile.cpp#L302

Can you add an exception for this, so it does not get deleted?

Slider0007:
> But I would not general skip deleting this file to avoid uncontrolled file size over time. I would rather only keep the file if time was not initially set at boot (which means some logs are in 1970-01-01 file). If during boot time is already set all infos are in correct dated file and regular deletion routine can delete all older files (including 1970-01-01).

I prepared it like described. Could you please test for verification.

caco3:
@Slider0007

On a cold start, I got the (unrelated) message
```
I (71939) TFLITE SERVER: Round #1 completed (58 seconds)
W (71999) TFLITE SERVER: Time server is configured, but time is not yet set. Check configuration
```

After some minutes it succeeded. maybe the "Check configuration" is a bit misleading as it is not a configuration issue.

And I like that you only delete it after a new start (with successful time)!

Slider0007:
> After some minutes it succeeded. maybe the "Check configuration" is a bit misleading as it is not a configuration issue.

It could be a config issue, but it's not for sure. Therefore, this part can can be deleted.

caco3:
> Therefore, this part can can be deleted.

Done

PR #2113, opened 2023-03-03
Use PSRAM for MQTT publishing and make it scheduled

caco3:
This PR replaces https://github.com/jomjol/AI-on-the-edge-device/pull/2112

- Make sure the initial sending of the Homeassistant Discovery and the Static Topics get done within the MQTT step instead of whenever Wifi got connected. This way we can avoid to have it consuming a lot of heap at the wrong time.
- When the publishing of the Homeassistant Discovery and the Static Topics get triggered by the REST API, only publish them in the MQTT step of the current round
- Allow setting QOS for MQTT publishing
- Add logging of heap change on MQTT publishing
- Use PSRAM for the MQTT Outbox, see CONFIG_MQTT_CUSTOM_OUTBOX and `code/components/jomjol_mqtt/mqtt_outbox.c`. This lowers the heap usage for the Homeassistant Discovery from `>20kBytes` to `<10kBytes`, see below.

## Not implemented
The following concepts did not get implemented as they no longer are needed:
- ~~Wait for MQTT transmission timeout after publishing with `QOS>0` to make sure the messages got sent or expired~~ 
   --> No longer needed as we have now enough heap
- ~~Change the MQTT QOS1 expiration time from (default) `30` to `5s`~~ 
   --> No longer needed as we have now enough heap
- ~~Use `QOS0` for Homeassistant Discovery topics. the messages then could possibly get lost but we save a lot of heap~~ 
   --> No longer needed as we have now enough heap

# Usage of PSRAM for the MQTT Outbox
Using QOS1 for all topics
PSRAM saves around 10kBytes on Heap!

## Using internal memory (normal case)
```
I (22101) MQTT SERVER: Int. Heap Usage before Publishing Homeassistand Discovery Topics: 41275, after: 20359, delta: 20916, lowest free: 11751
I (28311) MQTT SERVER: Int. Heap Usage before Publishing Static Topics: 20359, after: 19787, delta: 572, lowest free: 11751
I (34541) MQTT SERVER: Int. Heap Usage before publishing System Topics: 19787, after: 19275, delta: 512, lowest free: 11751
```

## Using PSRAM for the MQTT Outbox:
```
I (22831) MQTT SERVER: Int. Heap Usage before Publishing Homeassistand Discovery Topics: 41287, after: 31887, delta: 9400, lowest free: 21003
I (29031) MQTT SERVER: Int. Heap Usage before Publishing Static Topics: 31887, after: 31651, delta: 236, lowest free: 21003
I (35251) MQTT SERVER: Int. Heap Usage before publishing System Topics: 31651, after: 31547, delta: 104, lowest free: 21003
```


Slider0007:
@caco3: I've just flashed the firmware for testing. I couldn't find any issue with mqtt message sending. So this works perfect, even with 10kB less heap :-) I would keep the mqtt outbox in PSRAM.

The only thing I realized is that the debug logs of the early mqtt messages, in our case, connection, status and some discovery messages are not logged anymore. Logging starts with message id 15, Ihe IDs before are missing (even there are all sent). It's not that critical but I'd like to mention.
image

caco3:
hmm, thats odd.
on my system I see them all:
```

[0d00h00m12s] 2023-03-03T22:15:29	<INF>	[MQTT IF] Init
[0d00h00m12s] 2023-03-03T22:15:29	<INF>	[MQTT IF] Client started, waiting for established connection...
<DBG>	[MQTT IF] topic Testzaehler/ctrl/flow_start subscribe successful, msg_id=1
```

I wouldn't worry. Also it does not make sense that it skips some log writes.

caco3:
@jomjol @Slider0007 please test

for me we can merge it

caco3:
I rebased it, please test and merge

Slider0007:
@caco3: Please see comments above...

caco3:
description updated

Slider0007:
> description updated

What do you think about this textual adaption?

image

PS: It seems this review feature is not really helpful, because not obious enough...

caco3:
> PS: It seems this review feature is not really helpful, because not obious enough...

It is because you have only created the comment but did not submit it yet (its "pending".
You have to complete the review:
grafik

But I agree, its easy to miss.

Slider0007:
> > PS: It seems this review feature is not really helpful, because not obious enough...
> 
> It is because you have only created the comment but did not submit it yet (its "pending".
> You have to complete the review:
> grafik
> 
> But I agree, its easy to miss.

Thanks for your explanation. Sorry in this case my miss. Learning every day :-)

code/components/jomjol_mqtt/server_mqtt.cpp
```diff
@@ -191,26 +203,36 @@ bool publishSystemData() {
 
     LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "Publishing system MQTT topics...");
```
caco3, line 204:
```suggestion
    LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "Publishing System MQTT topics...");
```

code/components/jomjol_mqtt/server_mqtt.cpp
```diff
@@ -144,16 +146,18 @@ bool MQTThomeassistantDiscovery() {
 
     LogFile.WriteToFile(ESP_LOG_INFO, TAG, "MQTT - Sending Homeassistant Discovery Topics (Meter Type: " + meterType + ", Value Unit: " + valueUnit + " , Rate Unit: " + rateUnit + ")...");
```
caco3, line 147:
```suggestion
    LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "Publishing Homeassistant Discovery topics (Meter Type: '" + meterType + "', Value Unit: '" + valueUnit + "' , Rate Unit: '" + rateUnit + "') ...");
```

caco3:
No problem!

Did you write more than one comment? Because I still can not see it. But I updated the code according your suggestion.

Slider0007 (approved):

Slider0007:
It was just the one comment

PR #2114, opened 2023-03-03
Add GJ (gigajoule) as an energy meter unit

adarazs:
Some district heating meters are displaying their values in GJ and Home Assistant is also capable of accepting heating values in this format.

I added the rate as "GJ/h", even though `1 joule = 1 Ws`, so the correct rate would be `GW/s`. However it would be a silly small unit, so I think it would make more sense to use GJ/h. Let me know if I should change it.

I was able to use a sensor that had the unit of "GJ" in the HA energy dashboard by converting the value with a template sensor, but I think it is much cleaner to get the correct unit straight from the "AI-on-the-edge-device".

Testing: I did not compile and test this, just found the places where "kwh" was mentioned and added GJ there as well. Let me know if you need full testing before and I will set up a dev environment.

Thank you for this awesome project! I just tried to set it up the first time and it worked very well so far!

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowMQTT.cpp
@@ -156,6 +156,9 @@ bool ClassFlowMQTT::ReadParameter(FILE* pfile, string& aktparamgraph)
             else if (toUpper(splitted[1]) == "ENERGY_MWH") {
                 mqttServer_setMeterType("energy", "MWh", "h", "MW");
             }
+            else if (toUpper(splitted[1]) == "ENERGY_GJ") {
+                mqttServer_setMeterType("energy", "GJ", "h", "GJ/h");
+            }
         }
 
         if ((toUpper(splitted[0]) == "CLIENTID") && (splitted.size() > 1))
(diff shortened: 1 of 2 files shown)


caco3:
@adarazs Thanks for your contribution! It looks ok for me. Before I merge it, can you test it? You can use the build from https://github.com/jomjol/AI-on-the-edge-device/actions/workflows/build.yaml?query=branch%3Aadd_gigajoule

Just take the most top build, then scroll down and use the "update" zip as normal.
since this is based on rolling, you should make sure to be on v15.0 beforehand. thats also the oldest version you will be able to revert to afterwards.

adarazs:
Hi @caco3! Very cool build system, it was a breeze to test this.

It works as expected. I installed the version, the new option appears in the options, the unit changed correctly to GJ in Home Assistant and I could add it to the Energy Dashboard. The dashboard converts it to kWh gas, but it's correct after the conversion (0.02 GW usage showed up as 5.56 kWh).

Screenshots with the tested version:

gigajoule-1
--
gigajoule-2
--
gigajoule-3

caco3:
Nice, thanks!

And yes, HA does some automatic units conversion.

Issue #2092, opened 2023-02-27
Wifi Roaming not enabled yet 


caco3:
Support for Wifi Roaming is implemented, but it seems not enabled yet.

caco3:
@Slider0007 Ich war mal so frei und habe dich assigned, da du am ehesten weist, was noch fehlt.

@jomjol FYI

caco3:
https://github.com/jomjol/AI-on-the-edge-device/issues/1969

Slider0007:
No problem :-)

I've already started with heap analysis, but right now it looks not really promising without having risk of instability. One chance would be to reduce heap which is used by MQTT discovery topics. This would improve the situation --> see https://github.com/jomjol/AI-on-the-edge-device/pull/2091#issuecomment-1446837008

caco3:
As a reply to https://github.com/jomjol/AI-on-the-edge-device/pull/2091#issuecomment-1446837008:

I agree that we only should send any MQTT topics at a defined point in time.
REST API trigger or re-connecting of broker or wifi just should set a flag, but the sending should only happen at the end of the next round.

Do we understand why roaming requires so much additional memory? Can we adjust this?
Keep in mind that the models also will grow over time, thus will take more memory. I think there was already an issue that a model was too large.

As for HA Discovery, I still do not understand why it takes so much memory. In my understanding, MQTT publishing is a blocking task, thus it does not matter if I want to send 1 or 20 topics, the memory usage should be the same.

We might also consider to go and use the upper 4 Mbytes (Hi-Mem, I think @nliaudat looked into this, but I can't find the Github issue/PR) for roaming, eg. make it a requirement to have 8 MBytes for Wifi Roaming.
Most users do not need roaming (since they have only one Wifi Accesspoint), so for them 4 MB would still suffice. 
And the ones that need the roaming (and possibly other large features in the future), need to provide a 8MB device.
I ordered 3 times an ESP32-CAM, and it always came with 8 MB (untested, but they are shown as 8 MB). So I am not surer if that many 4 MB devices are around.

Slider0007:
> I agree that we only should send any MQTT topics at a defined point in time.
> REST API trigger or re-connecting of broker or wifi just should set a flag, but the sending should only happen at the end of the next round.

Would it not be easier to just trigger by button which is interlocked with flow finished?

> Do we understand why roaming requires so much additional memory? Can we adjust this?

[...]

PR #2120, opened 2023-03-04
"WIFI roaming" by channel scan (AP switching at low RSSI)

Slider0007:
Implement a kind of roaming feature #2092 which has less dependencies in terms of capabilities of AP / mesh system and is lightwight to ensure less impact to memory usage of ESP32.

# WIFI AP switching by channel scan (simplified "roaming")
- Implement wifi channel scanning (background scan, scan possible even ESP is connected) to have kind of "roaming" like feature
- Function needs to be activated using RSSIThreshold parameter by setting a value lower than 0 (0 = disable)
- Scan and evaluation gets triggered when actual RSSI < RSSI Threshold (RSSIThreshold, e.g. -75dBm) and flow is finished (Info gets printed with DEBUG level)
image
- Process:
 - Scan all channels for configured SSID
 - Compare if (actual RSSI + 5) < AP RSSIs found by scan (+ 5 to avoid switching by almost equal RSSI)
 - If an AP with better RSSI is found, force wifi disconnect to reconnect to AP with hopefully better RSSI
 - The reconnect to new AP will not get forced by BSSID, to avoid issues when potentially AP with better RSSI reject connection for some reason. Therefore the selection is related to the following config and wifi library.
image
- Switch on all channel scan option --> Wifi scan will performed for all channels before choosing an AP and not use first matching AP (This is also valid even AP switching function is disabled => RSSIThreshold = 0)

# Alternative: WIFI roaming 802.11kv --> deactivated by default
NOTE: Fully implemented, but not fully tested (be aware: additional ca. 6-10kB internal heap is consumed to run this feature)

- basic roaming functionality is disabled by default, compile time (--> This feature needs to be activated in define.h --> #define WLAN_USE_MESH_ROAMING and sdkconfig.default + disabling roaming by scanning feature)
- This is needed to handle any roaming request triggered by AP/mesh (802.11k, 802.11v)
- Optional WIFI roaming feature (needs to be enabled in defines.h --> #define WLAN_USE_MESH_ROAMING_ACTIVATE_CLIENT_TRIGGERED_QUERIES) to enable ESP triggered query to roam using RSSIThreshold parameter can be activated when parameter is set to a value lower than 0 (0 = disable this optional feature)
--> The query is only sent when actual RSSI < RSSI Threshold (RSSIThreshold, e.g. -75dBm)
--> The evaluation is only triggered after flow is finished and RSSI < RSSI Threshold every round (The trigger and the neighbour table gets printed with DEBUG level)
image

--> This does not guarantee that ESP will roam because only a query to AP/mesh is going to be sent. The actual request that ESP needs to roam is still triggered by AP/mesh. If AP/mesh does not see any need to roam, no roam request will sent to ESP and nothing will happen

The info if AP is supporting the roaming features 802.11k and 802.11v will be logged to logfile after wifi connection gets established
image

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowControll.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowControll.cpp
@@ -612,8 +612,8 @@ bool ClassFlowControll::ReadParameter(FILE* pfile, string& aktparamgraph)
 
         /* TimeServer and TimeZone got already read from the config, see setupTime () */
         
-        #if (defined WLAN_USE_MESH_ROAMING && defined WLAN_USE_MESH_ROAMING_ACTIVATE_CLIENT_TRIGGERED_QUERIES)
-        if (((toUpper(splitted[0]) == "RSSITHRESHOLD") || (toUpper(splitted[0]) == "RSSITHREASHOLD")) && (splitted.size() > 1)) // Workaround for typo RSSITHREASHOLD
+        #if (defined WLAN_USE_ROAMING_BY_SCANNING || (defined WLAN_USE_MESH_ROAMING && defined WLAN_USE_MESH_ROAMING_ACTIVATE_CLIENT_TRIGGERED_QUERIES))
+        if ((toUpper(splitted[0]) == "RSSITHRESHOLD") && (splitted.size() > 1))
         {
             int RSSIThresholdTMP = atoi(splitted[1].c_str());
             RSSIThresholdTMP = min(0, max(-100, RSSIThresholdTMP)); // Verify input limits (-100 - 0)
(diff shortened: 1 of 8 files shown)


caco3:
general question:
With the threshold set to 0, what will happen on the startup? Does it still use the first AP it finds or does it a scan und picks the best one? I would prefer the 2nd option, so also users who do not want to go into expert functions get this improvement. Most might even be unaware or not able to understand what it does.

Slider0007:
> With the threshold set to 0, what will happen on the startup? Does it still use the first AP it finds or does it a scan und picks the best one? I would prefer the 2nd option, so also users who do not want to go into expert functions get this improvement. Most might even be unaware or not able to understand what it does.

The threshold only activates the evaluation to switch APs. If set to 0 no triggered switching will happen.

After any scan always best AP is / should be selected. This a wifi library config setting:
image

caco3:
so the picking of the best AP on startup works also with threshold=0?

Slider0007:
> so the picking of the best AP on startup works also with threshold=0?

yes, as long the wifi library is working correctly

PR #2130, opened 2023-03-05
Update FeatureRequest.md

Yveaux:
Added #36 Run demo without camera


caco3:
I will merge it, but I don't think somebody will implement it.
Reason is that we want to keep the timing as close to normal operation as possible, so it can be used for testing.

Issue #2168, opened 2023-03-12
Timezone not properly read from config


caco3:
### Discussed in https://github.com/jomjol/AI-on-the-edge-device/discussions/2166

<div type='discussions-op-text'>

<sup>Originally posted by **FastFrank** March 12, 2023</sup>
### The Problem

Timezone in Configuration:

CET-1CEST,M3.5.0,M10.5.0/3

Log from MQTT last night: (12-03-2023)

JSON gewijzigd in { "value": "18839.45", "raw": "18839.45", "pre": "18839.45", "error": "no error", "rate": "0.008000", "timestamp": "2023-03-12T03:04:41+0200" }
02:06:01 - 8 uur geleden
JSON gewijzigd in { "value": "18839.41", "raw": "18839.41", "pre": "18839.41", "error": "no error", "rate": "0.004000", "timestamp": "2023-03-12T01:59:41+0100" }
02:01:00 - 8 uur geleden

### Version

15.0.3

### Logfile

```shell
[0d00h00m04s] 2023-03-12T08:52:26	<INF>	[SNTP] TimeServer: nl.pool.ntp.org
[0d00h00m04s] 2023-03-12T08:52:26	<INF>	[SNTP] Configuring NTP Client...
[0d00h00m04s] 2023-03-12T10:52:26	<INF>	[SNTP] Time zone set to CET-1CEST
[0d00h00m04s] 2023-03-12T10:52:26	<INF>	[SNTP] Time is already set: 2023-03-12 10:52:26
```

### Expected Behavior

Change the +1:00 to +02:00 on Month 3 last week day 0

in the log:
[SNTP] Time zone set to CET-1CEST,M3.5.0,M10.5.0/3
instead of:
[SNTP] Time zone set to CET-1CEST

### Additional Context

_No response_</div>

FastFrank:
I was looking in other issues like #2150 and there the log was:

[0d00h00m04s] 2023-03-08T19:48:23 <INF> [SNTP] TimeZone not set, using default: CET-1CEST,M3.5.0,M10.5.0/3
[0d00h00m04s] 2023-03-08T19:48:23 <INF> [SNTP] Configuring NTP Client...
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [SNTP] Time zone set to CET-1CEST,M3.5.0,M10.5.0/3
[...]

PR #2169, opened 2023-03-12
Fix timezone config parser

caco3:
Solves https://github.com/jomjol/AI-on-the-edge-device/discussions/2166 resp. https://github.com/jomjol/AI-on-the-edge-device/issues/2168


caco3:
@jomjol @Slider0007 FYI, the Timezone value contains a `,`, but the parser splitted the string there.

Issue #2200, opened 2023-03-16
Out of memory


wg25:
### The Problem

Seit der Version 15.1.0 habe ich nach kurzer Laufzeit folgende Fehlermeldungen im LOG:

[0d09h18m50s] 2023-03-16T17:34:39 <ERR> [TFLITE] CTfLiteClass::ReadFileToModel: Can't allocate enough memory: 1200904
[0d09h18m50s] 2023-03-16T17:34:39 <ERR> [CNN] Can't load tflite model /config/dig-cont_0600_s3.tflite -> Exec aborted this round!

Restarts wurden mehrfach durchgeführt. Das Problem ist bereits auf zwei ESP32 aufgetreten.

Bilder werden gemacht, aber die Digitalisierung bricht wegen Speichermangel ab.

### Version

Release: v15.1.0 (Commit: 7d33c3e+)

### Logfile

```shell
[0d00h00m04s] 2023-03-16T07:14:21 <INF> [MAIN] =================================================
[0d00h00m04s] 2023-03-16T07:14:21 <INF> [MAIN] ==================== Start ======================
[0d00h00m04s] 2023-03-16T07:14:21 <INF> [MAIN] =================================================
[0d00h00m04s] 2023-03-16T07:14:21 <INF> [SDCARD] Basic R/W check started...
[0d00h00m04s] 2023-03-16T07:14:21 <INF> [SDCARD] Basic R/W check successful
[0d00h00m04s] 2023-03-16T07:14:22 <INF> [SNTP] TimeServer not defined, using default: pool.ntp.org
[0d00h00m05s] 2023-03-16T07:14:22 <INF> [SNTP] Configuring NTP Client...
[0d00h00m05s] 2023-03-16T08:14:22 <INF> [SNTP] Time zone set to CET-1CEST,M3.5.0,M10.5.0/3
[0d00h00m05s] 2023-03-16T08:14:22 <INF> [SNTP] Time is already set: 2023-03-16 08:14:22
[0d00h00m05s] 2023-03-16T08:14:22 <INF> [MAIN] CPU frequency: 160 MHz
[0d00h00m06s] 2023-03-16T08:14:23 <INF> [SDCARD] Folder/file presence check started...
[0d00h00m06s] 2023-03-16T08:14:23 <INF> [SDCARD] Folder/file presence check successful
[...]

PR #2206, opened 2023-03-17
Show PSRAM usage

caco3:
- Wrap all application driven PSRAM alloc/free calls into common functions and log them.
    Notes:
     -  This excludes the PSRAM usage of the MQTT Outbox as this is not part of the application.
     - Any 3th party library using PSRAM is not added to this wrappers
- Fix use-after-free (https://github.com/jomjol/AI-on-the-edge-device/pull/2206/commits/1e6aa3f75117982f64bbc4346b86c7b66b14ddbf)
- ~~Free `rgb_image` before re-using it (https://github.com/jomjol/AI-on-the-edge-device/pull/2206/commits/79e0911c4ce4e09e0c89c9c20446158df78b6378)~~
- add a name parameter to all CImageBasis for clearer logging

Investigations:
image


caco3:
@Slider0007 @jomjol 

I made this changes to have a log line on every PSRAM alloc/free. It is a preparation for static allocations to avoid issues like in https://github.com/jomjol/AI-on-the-edge-device/issues/2200

Below I colorized all alloc/free calls (Startup until end of first round):
image

Filtered by the variables:
image
=> **OK**

image
=> OK

image
=> **NOT OK**
    It seems as if `rgb_image` alsways `NULL`, so no `free` is needed. But I do not understand this!

code/components/jomjol_image_proc/CImageBasis.cpp
```diff
@@ -470,7 +477,12 @@ CImageBasis::CImageBasis(CImageBasis *_copyfrom)
     #endif
 
     int memsize = width * height * channels;
-    rgb_image = (unsigned char*)GET_MEMORY(memsize);
+
+    if (rgb_image != NULL) {
+        free_psram_heap(std::string(TAG) + "->rgb_image (CImageBasis1)", rgb_image);
```
Slider0007, line 482: It doesn't make sense to me to free a for sure non created oject variable in a constructor. This a class therefore you have to think in objects and not in variables. Objects get created using constructor and needs to be deleted after usage on object level. This happens calling delete "object". Then Destructor gets called (in desctrutor, object variables needs to be freed then).
This comments is also valid for all other constructors in this class.

code/components/jomjol_tfliteclass/CTfLiteClass.cpp
```diff
@@ -305,17 +306,17 @@ CTfLiteClass::CTfLiteClass()
...
 
 CTfLiteClass::~CTfLiteClass()
 {
-  free(modelfile);
-
-  free(this->tensor_arena);
   delete this->interpreter;
   delete this->error_reporter;
+
+  free_psram_heap(std::string(TAG) + "->modelfile", modelfile);
```
Slider0007, line 318: There is no difference if you call free or caps_malloc_free even memory is allocated in PSRAM.

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html
Because malloc() uses the capabilities-based allocation system, memory allocated using [heap_caps_malloc()](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html#_CPPv416heap_caps_malloc6size_t8uint32_t) can be freed by calling the standard free() function.

code/components/jomjol_image_proc/CRotateImage.cpp
```diff
@@ -336,7 +339,7 @@ void CRotateImage::Translate(int _dx, int _dy)
     memCopy(odata, rgb_image, memsize);
     if (!ImageTMP)
     {
-        stbi_image_free(odata);
```
Slider0007, line 339: Have you checked if in the function stbi_image_free something is handled special that you can repalce it?

code/components/jomjol_image_proc/CImageBasis.cpp
```diff
@@ -572,6 +593,11 @@ bool CImageBasis::ImageOkay(){
 CImageBasis::CImageBasis(uint8_t* _rgb_image, int _channels, int _width, int _height, int _bpp)
 {
     islocked = false;
+
+    if (rgb_image != NULL) {
+        free_psram_heap(std::string(TAG) + "->rgb_image (CImageBasis4)", rgb_image);
```
Slider0007, line 598: Here it's not even memory allocated. This is just pointer hand-over

code/components/jomjol_image_proc/CImageBasis.cpp
```diff
@@ -608,7 +634,7 @@ CImageBasis::~CImageBasis()
     RGBImageLock();
 
     if (!externalImage)
-        stbi_image_free(rgb_image);
+        free_psram_heap(std::string(TAG) + "->rgb_image (~CImageBasis)", rgb_image);
```
Slider0007, line 637: Have you checked if in the function stbi_image_free something is handled special that you can repalce it?
Comment valid for all stbi_image_free usages

Slider0007 (changes requested):

Slider0007:
> #### rgb_image
> image
> 
> => **NOT OK** It seems as if `rgb_image` always `NULL`, so no `free` is needed. But I do not understand this! Normally I would expect a memory leak because of this but this seems not to be the case!


image

You have to think in objects and not in variables, because a class is used to handle CBasisImage objects.
Here all ROIs objects are get created at init. The are not deleted and recreated every round, because they are used all the time to display ROI infos. For this reason the constructor is not called again (so your "free" in constructor is also not called anymore)

I think we can see this as kind of static because allocated early state in a deterministic way. I wouldn't do anything these objects (they are also limited in size).

#######################
I would remove all the frees from any constructors.
Whenever a object is not used aynmore it needs to be deleted. Call the delete command, destructor of class is called and object variables needs to be freed in there. Therefore you do not see a memory leakage of CBasisImage, because this is already fully implemented and working.
As an example: CImageTMP in Alginment Class
Object get created
https://github.com/jomjol/AI-on-the-edge-device/blob/a8d7b29507312c792ff438b4183c4295a0b41619/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp#L192

Then use it and delete after usage:
https://github.com/jomjol/AI-on-the-edge-device/blob/a8d7b29507312c792ff438b4183c4295a0b41619/code/components/jomjol_flowcontroll/ClassFlowAlignment.cpp#L271

With this deleteion call, also rgb_image as an object variable get freed like programmed in class destructor

Slider0007:
> * Fix use-after-free ([1e6aa3f](https://github.com/jomjol/AI-on-the-edge-device/commit/1e6aa3f75117982f64bbc4346b86c7b66b14ddbf))

What do you mean with this?

caco3:
> > * Fix use-after-free ([1e6aa3f](https://github.com/jomjol/AI-on-the-edge-device/commit/1e6aa3f75117982f64bbc4346b86c7b66b14ddbf))
> 
> What do you mean with this?

After I replaces `free()` with the correct `free_psram_heap()`, it crashed on the delete. Thats to be expected because the allocated memory is not available anymore. The delete must come before.

N.B. all free calls are ok with providing them a NULL pointer, so redundant free calls are ok.

code/components/jomjol_tfliteclass/CTfLiteClass.cpp
caco3: yes, that is what i also would expect, but my experience below (https://github.com/jomjol/AI-on-the-edge-device/pull/2206#issuecomment-1474788761) seems to prove different!

code/components/jomjol_image_proc/CImageBasis.cpp
caco3: right, I will remove them again in the constructor.

caco3:
@Slider0007 Thanks for your feedback. I will remove the changes regarding the `rgb_image` again, but only at the end of the weekend as I am busy now.

Slider0007:
> After I replaces `free()` with the correct `free_psram_heap()`, it crashed on the delete. Thats to be expected because the allocated memory is not available anymore. The delete must come before.

This I do not fully understand because the four elements are four independed objects/memory areas, but as long it works... :-)

caco3:
@Slider0007 

In todays commits I changed following:
 - reverted freeing where not needed (as discussed)
 - add a name parameter to all CImageBasis for clearer logging

### Outcome:
image

I would like to merge this PR as is and create a separate PR where we tune the memory allocation.
This MR should have no function changes beside the changes needed for the debug logging.

caco3:
Having a look on the larger memory users, i don^t think we are short of PSRAM. It is more likely as you say a problem of fragmentation.
To work around this, we would need to keep some memory allocated and reuse it, eg. `ImageTMP` and `tensor_arena`

Slider0007 (approved):

jomjol:
> Having a look on the larger memory users, i don^t think we are short of PSRAM. It is more likely as you say a problem of fragmentation.
> To work around this, we would need to keep some memory allocated and reuse it, eg. `ImageTMP` and `tensor_arena`

I totally agree. There are 3 major memory consumers:
tflite:
1) TFlite for internal processing
```
    this->kTensorArenaSize = 800 * 1024;   /// 800k
    this->tensor_arena = (uint8_t*)GET_MEMORY(kTensorArenaSize);
```
2) Camera image (640x480*3 = 921k)
3) internal memory for image processing (especially alignment and rotation) - reuse memory via `ImageTMP`

All of this should is with a sum of 2642 kByte in the PSRAM and should not be a problem. There are some other smaller objects in there, but all together should fit quite well. If we shift other topics (like MQTT) into the PSRAM, we might run into a problem with fragmentation.
We can solve this by reserving the memory as one of the first steps.

Slider0007:
As stated already here there are a lot more larger consumers: https://github.com/jomjol/AI-on-the-edge-device/issues/2200#issuecomment-1474783093

Biggest consumers of PSRAM
```
CAM buffer ca. 65kB --> needed all the time, mandatory
CImageBasis: 920kB --> needed all the time
ALG_ROI JPG 128KB --> needed all the time, mandatory to show image
ROI images (ca. 165kB as @caco3 stated in his eval) --> needed all the time, mandatory to show ROI info
WIFI / Network stack (ca. ?kB) --> needed all the time
--> All allocated during boot or init --> kind of static, no delete --> so in my opionion not critical

MQTT outbox (ca. 10kB?) --> unsure, because it depends when WIFI connection is ready and MQTT service gets started, maybe creates fragmentation, but this is just guessing

Share PSRAM to avoid having to much allocated at the same time
CImageTMP: 920KB --> only needed in alignment
Tensor memory ca. 820kB --> only needed in ROI processing
Model up to 1200kb --> only needed in ROI processing
```

Eaxmaple from my eval:
```
Showing data for heap: 0x3f800000
Block 0x3f80086c data, size: 40 bytes, Free: No
Block 0x3f800898 data, size: 20 bytes, Free: No
Block 0x3f8008b0 data, size: 61440 bytes, Free: No --> CAM
Block 0x3f80f8b4 data, size: 55880 bytes, Free: No --> WIFI?
Block 0x3f81d300 data, size: 921600 bytes, Free: No --> BasisImage
Block 0x3f8fe304 data, size: 128004 bytes, Free: No --> ALGROI JPG
Block 0x3f91d70c data, size: 1920 bytes, Free: No -> From here ROIs
Block 0x3f91de90 data, size: 4860 bytes, Free: No
Block 0x3f91f190 data, size: 1920 bytes, Free: No
Block 0x3f91f914 data, size: 4860 bytes, Free: No
Block 0x3f920c14 data, size: 1920 bytes, Free: No
Block 0x3f921398 data, size: 4860 bytes, Free: No
Block 0x3f922698 data, size: 3072 bytes, Free: No
Block 0x3f92329c data, size: 25392 bytes, Free: No
Block 0x3f9295d0 data, size: 3072 bytes, Free: No
Block 0x3f92a1d4 data, size: 25392 bytes, Free: No
Block 0x3f930508 data, size: 3072 bytes, Free: No
Block 0x3f93110c data, size: 25392 bytes, Free: No
Block 0x3f937440 data, size: 3072 bytes, Free: No
Block 0x3f938044 data, size: 25392 bytes, Free: No -> ROIs
--> till here it's kind of static because not delted anymore

Block 0x3f93e378 data, size: 2890880 bytes, Free: Yes
--> The rest is the "breathing part" where ImageTMP, tensor and model need to fit
--> ca. 920kB + 820kB, 1200kB = 2940kB
--> so would not fit if we just allocate all of them at the beginning
```

Right now ImageTMP gets freed before tensor and model gets loaded --> there no issue right now (if we do not have any fragmentation)

caco3:
we could have memory allocation large enough for `tensor_arena` and `model` together (approx `2 MBytes` plus some spare) and then re-use it for `ImageTMP`. This way we would use it more efficient.

As for the fragmentation:
the MQTT Outbox has 2 places where it allocates RAM:
1. `outbox_init()` -> at init, kind of static
2. `outbox_enqueue()` -> most likely at every publishing -> most likely causing the fragmentation.

Slider0007:
> we could have memory allocation large enough for `tensor_arena` and `model` together (approx `2 MBytes` plus some spare) and then re-use it for `ImageTMP`. This way we would use it more efficient.
> 
> As for the fragmentation:
> the MQTT Outbox has 2 places where it allocates RAM:
> 1. `outbox_init()` -> at init, kind of static
> 2. `outbox_enqueue()` -> at every publishing -> most likely causing the fragmentation.
> 

I agree, I also assume fragmentation has something to do with MQTT message transmission. And this is unfortunately not deterministic. This is the reason I suggested to remove this from PSRAM, as long we do not have any other better solution.

PR #2209, opened 2023-03-19
Reboot task - increase waiting time

Slider0007:
For some reason since 15.1 and newer reboot logic is not working reliable anymore for me. It takes sometimes way longer to handle e.g. MQTT disconnect / MQTT destroy. 

Sometimes it's not rebooting at all, but sometimes it's still working --> Increase waiting time to give single shutdown tasks more time to complete.

@caco3, @jomjol: Can you please check this topic on your side.


Slider0007:
My today's test with latest rolling also do not show this issue anymore. I cannot tell what caused this the last time... Therefore I close this PR for the moment.

PR #2215, opened 2023-03-20
Static shared memory in PSRAM for model/imageTMP and tensor_arena

caco3:
# Proof of Concept
:bangbang: Do not merge!

## Use same memory block for all `tensor_area` 
This is working ok

## Use same memory block for both models and `imageTMP`.
If the allocated memory gets too small, it gets freed an a larger block gets allocated. This happens withing the first round and then no change is needed anymore.

This seems to work fine for the first round but at the 2nd round I always run into
```
I (24722) C IMG BASIS: Image (zwImage) loaded from memory: 0, 0, 0
E (24722) C IMG BASIS: Image with size 0 loaded --> reboot to be done! Check that your camera module is workin
```

The `fb` how ever contains valid data!

PR #2117, opened 2023-03-03
move Wifi, LWIP and BSSI to PSRAM

caco3:
This saves us around 30kBytes of heap!

Some tests I run (same config), chcked after 3 rounds:
- rolling (39b8b5b): Int Free: 33159, Int Min Free: 10007
- With `CONFIG_SPIRAM_TRY_ALLOCATE_WIFI_LWIP` and `CONFIG_SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY`: Int Free: 69731, Int Min Free: 63571
@needs some more testing, but for me I do not see any issue as of now.

PR #2224, opened 2023-03-22
Revert PSRAM usage as it lead to memory fragmentation.

caco3:
In https://github.com/jomjol/AI-on-the-edge-device/pull/2117/files we moved Wifi, LWIP and BSSI to PSRAM.
This freed a lot of internal RAM. How ever as found in https://github.com/jomjol/AI-on-the-edge-device/issues/2200 and further analyzed in https://github.com/jomjol/AI-on-the-edge-device/pull/2215 we found out that the changes cause additional memory fragmentation. This makes it impossible to allocate the huge memory blocks as needed for the `model` (up to 1.2 MB), `tensor_arena` (currently around 800 kB) and `imageTMP` (around 900 kB).

To get back to a stable system, we revert the moves to PSRAM for now

### DIFF
--- a/code/sdkconfig.defaults
+++ b/code/sdkconfig.defaults
@@ -100,8 +100,8 @@ CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL=16384
 CONFIG_SPIRAM_MALLOC_RESERVE_INTERNAL=40960
 CONFIG_SPIRAM_CACHE_WORKAROUND=y
 CONFIG_SPIRAM_IGNORE_NOTFOUND=y
-CONFIG_SPIRAM_TRY_ALLOCATE_WIFI_LWIP=y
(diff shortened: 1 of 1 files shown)


caco3:
@Slider0007 @jomjol 

- [ ] Will this get us back to being short of internal RAM, eg. during Homeasistant Discovery sending?
- [ ] Shall we keep the MQTT Outbox in PSRAM? Since its memory usage is in our hands, we could change it to static allocations. 
   - https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_mqtt/mqtt_outbox.c#L37
 ist most likely static and run at startup
   - https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_mqtt/mqtt_outbox.c#L49 and https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/components/jomjol_mqtt/mqtt_outbox.c#L61 seem to by dynamic, but we could change this to a large enough static block.

PR #2233, opened 2023-03-23
Remove trailing slash in influxDBv1

caco3:


srLinux:
ksnip_20230323-233805
yip, now it is working! thank you :+1:

PR #2278, opened 2023-04-01
Fix time offset issues in InfluxDB component.

### LINKED ISSUE #2150: Incorrect handling of timezone in InfluxDB component
### The Problem

It looks like since a9c5bebb, the timestamps of the data points sent to influxdb (at least with the v1 API) are wrong.

Running a device in the CET timezone, properly configured (so that the logs and web interface correctly show the local time), the data points are written two hours back in time, whereas before they used to be added with the correct time.

I am not sure what motivated a9c5bebb, because the time zone should not have any influence in this interface: just using the UTC time and sending it as such to InfluxDB worked well before. @jomjol could you explain why you made this change?

I am a bit puzzled by this time arithmetic:
https://github.com/jomjol/AI-on-the-edge-device/blob/dc9f5e597fc4ce2b89a27cc28f05553f0029f5bc/code/components/jomjol_influxdb/interface_influxdb.cpp#L56-L63

### Version

15.0.3

### Logfile

```shell
[0d00h00m04s] 2023-03-08T19:48:23 <INF> [SNTP] TimeServer: 192.168.5.2
[0d00h00m04s] 2023-03-08T19:48:23 <INF> [SNTP] TimeZone not set, using default: CET-1CEST,M3.5.0,M10.5.0/3
[0d00h00m04s] 2023-03-08T19:48:23 <INF> [SNTP] Configuring NTP Client...
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [SNTP] Time zone set to CET-1CEST,M3.5.0,M10.5.0/3
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [SNTP] Time is already set: 2023-03-08 20:48:23
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [MAIN] =================================================
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [MAIN] ==================== Startup ====================
[0d00h00m04s] 2023-03-08T20:48:23 <INF> [MAIN] =================================================
[0d00h00m04s] 2023-03-08T20:48:24 <INF> [MAIN] Release: v14.0.3 (Commit: 6c891ab+), Date/Time: 2023-02-05 19:36, Web UI: Release: v14.0.3 (Commit: 6c891ab+)
```

### Expected Behavior

Correct handling of time in the InfluxDB component.

### Screenshots

_No response_

### Additional Context

_No response_

### LINKED ISSUE #2273: weird timezone issues in influxdb
### The Problem

hi,
influxdb client always should send the UTC time. It causes weird issues like "in the future" measurements. Take a look over here: https://community.grafana.com/t/weird-timezone-problems/72815

ksnip_20230331-200845

After i changed the timezone on the webpanel to UTC0 the problem went away. It is maybe a influx/grafana related problem but the recommendation is to set the timezone to UTC in on the measurement sending clients. 

### Version

rolling 15.1.1

### Logfile

```shell
-
```

### Expected Behavior

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

wetneb:
Closes #2273.
Closes #2150.

This reverts back to using the UTC time to send timestamps to InfluxDB.

### DIFF
--- a/code/components/jomjol_influxdb/interface_influxdb.cpp
+++ b/code/components/jomjol_influxdb/interface_influxdb.cpp
@@ -46,33 +46,12 @@ void InfluxDB_V2_Publish(std::string _key, std::string _content, std::string _ti
 
     LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "InfluxDB_V2_Publish - Key: " + _key + ", Content: " + _content + ", Timestamp: " + _timestamp);
 
-    // Format:     #define PREVALUE_TIME_FORMAT_OUTPUT "%Y-%m-%dT%H:%M:%S%z"
-
+    time_t now = time(NULL);
     char nowTimestamp[21];
-    std::string payload;
-
-    if (_timestamp.length() > 0)
-    {
-        struct tm tm;
-        strptime(_timestamp.c_str(), PREVALUE_TIME_FORMAT_OUTPUT, &tm);
-        time_t t = mktime(&tm); // Time in Localtime (looks like timezone is not used by strptime)
-
-        struct tm * ptm;
-        ptm = gmtime ( &t );
-        time_t utc = mktime(ptm);
-        utc = 2*t - utc;        // Take care of timezone (looks difficult, but is easy: t = t + (t - utc), weil t-utc = timezone)
-
-        sprintf(nowTimestamp,"%ld000000000", (long) utc);           // UTC
-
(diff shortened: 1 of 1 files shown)


jomjol:
@wetneb : you choose here the timestamp directly out of the current time. But I want the time from the time the picture is taken, because this is the time, which is valid for the readout. I have merged your commt therefore to a dedicated branch.
There we can try to fix the problem within this repo: https://github.com/jomjol/AI-on-the-edge-device/tree/debug-influxdb

If you would like to support, please commit there. Thanks!

PR #2283, opened 2023-04-01
Debug influxdb

jomjol:
* working on the problem of influx db timeshift 

https://github.com/jomjol/AI-on-the-edge-device/issues/2273
https://github.com/jomjol/AI-on-the-edge-device/issues/2150


wetneb (commented):
Thank you for your explanation and for working on this problem!

I think the timestamp should not be passed around as a `std::string` between modules, but rather with something like a `time_t` for instance. That would remove the need to parse it back to a time_t in a situation like this (which was the source of the bug). The time should only then be converted to string when it is displayed in the UI, in the logs, or serialized somewhere.

The logic to write the InfluxDB line could also be shared between the two InfluxDB modules so that our changes do not have to be repeated twice.

PR #2286, opened 2023-04-03
Implement a camera livestream handler

Slider0007:
Implementation of a camera live stream REST API
--> /stream -> flashlight off
--> /stream?flashlight=true -> flashlight on
- Slow refresh rate to not stress the system, only 2 FPS
- Flow processing is still continue working in background (maybe a little slower)
- LIMITATION: Not using extra memory no additional dedicated stream webserver is implemented for this feature (reuse of web interface server in a kind of "blocking way"). This means either web interface is fully functional or the stream, but not both at the same time. But this is sufficient for the intended use case.

Follow up:
Implement this livestream API in the initial setup routine as step 1 to have ability to adjust focus and do initial lighting tests in a convient way.

### DIFF
--- a/code/components/jomjol_controlcamera/ClassControllCamera.cpp
+++ b/code/components/jomjol_controlcamera/ClassControllCamera.cpp
@@ -32,6 +32,14 @@
 
 static const char *TAG = "CAM"; 
 
+
+/* Camera live stream */
+#define PART_BOUNDARY "123456789000000000000987654321"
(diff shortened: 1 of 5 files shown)


lmdc45:
Thanks for this, actually had to flash https://github.com/easytarget/esp32-cam-webserver first to configure focus and appropriate distance. this will make it a lot easier

PR #2319, opened 2023-04-16
Extend InfluxDBv1 with individual topic names

jomjol:
In order to fully customize also the field values for InfluxDB V1, I have implemented there also numbers specific configurable field names. They now can be set independly for InfluxDB V1 and V2.

grafik


caco3:
did you also update the param docs?

caco3:
implements https://github.com/jomjol/AI-on-the-edge-device/issues/2297

PR #2326, opened 2023-04-17
analog ROI screen: Activate save button after ROI creation

### LINKED ISSUE #2263: Saving new analog number not possible
### The Problem

after adding a new analog number, the save button stays disabled (which is ok as there is no ROI in it).
how ever after adding a ROI the save button still stays disabled.

Note:
On the digits this works correctly

### Version

15.1.1

Slider0007:
fixes #2263

@caco3: Please test if this fixes your issue


caco3:
Thanks, I can confirm it is working now again.

Issue #2358, opened 2023-04-27
Loading theAlignment Marsk Page takes very long


caco3:
### The Feature

For some reason it loads a folder from the SD-Card for several times:
```
I (50303) OTA FILE: Deleting file: /img_tmp/ref0.jpg
I (50623) OTA FILE: Found file: main_dig1.bmp (1 KiB bytes)
I (50643) OTA FILE: Found file: leer.txt (0 B bytes)
I (50653) OTA FILE: Found file: main_dig2.bmp (1 KiB bytes)
I (50673) OTA FILE: Found file: main_dig3.bmp (1 KiB bytes)
I (50683) OTA FILE: Found file: main_ana1.bmp (3 KiB bytes)
I (50703) OTA FILE: Found file: main_ana2.bmp (3 KiB bytes)
I (50723) OTA FILE: Found file: main_ana3.bmp (3 KiB bytes)
I (50733) OTA FILE: Found file: main_ana4.bmp (3 KiB bytes)
I (50753) OTA FILE: Found file: ref0_org.jpg (3 KiB bytes)
I (50773) OTA FILE: Found file: ref1.jpg (1 KiB bytes)
I (50783) OTA FILE: Found file: ref1_org.jpg (1 KiB bytes)
I (50803) OTA FILE: Found file: main_dig1.jpg (1 KiB bytes)
I (50823) OTA FILE: Found file: main_dig2.jpg (1 KiB bytes)
I (50843) OTA FILE: Found file: main_dig3.jpg (1 KiB bytes)
I (50863) OTA FILE: Found file: main_ana1.jpg (1 KiB bytes)
I (50883) OTA FILE: Found file: main_ana2.jpg (1 KiB bytes)
I (50893) OTA FILE: Found file: main_ana3.jpg (1 KiB bytes)
I (50913) OTA FILE: Found file: main_ana4.jpg (1 KiB bytes)
I (50933) OTA FILE: Found file: ref0_org_org.jpg (3 KiB bytes)
I (50953) OTA FILE: Found file: ref1_org_org.jpg (3 KiB bytes)
I (50973) OTA FILE: Found file: name_name.jpg (627 B bytes)
I (50993) OTA FILE: Found file: main1_dig1.jpg (853 B bytes)
I (51013) OTA FILE: Found file: main1_dig2.jpg (840 B bytes)
I (51033) OTA FILE: Found file: main1_dig3.jpg (807 B bytes)
[...]

Issue #2347, opened 2023-04-24
Input sanitization missing for config "Decimal Shift" 


torwag:
The config to shift the decimal point (Decimal Shift) after recognition of all digits, expects the number of digits to move the decimal point. E.g. 1376 l are 1.376 m3 thus -3 digits.

Obvious my overtired engineer brain did what it always does and divided by 1000 and as this, I entered 1000 into that field.

As a side effect that seems to expose a missing input sanitization. The ESP simply silently crashes. It might be a good idea to limit inputs between -9 and 9 as everything else seems rather unlikely.

_Originally posted by @torwag in https://github.com/jomjol/AI-on-the-edge-device/discussions/2340#discussioncomment-5702989_

caco3:
this is an issue we have with almost any input parameters

Issue #2225, opened 2023-03-22
Feature Request: Mobile Device Webpage


caco3:
### Discussed in https://github.com/jomjol/AI-on-the-edge-device/discussions/2219

<div type='discussions-op-text'>

<sup>Originally posted by **8chM** March 22, 2023</sup>
It is possible to optimize the webpages, for mobile devices?
Of course it is very hard, to configure the ROIs with an iphone, but for overview it will be more helpfull.
</div>

## New Webpage (rolling)
- [x] Stream should open in new page
- [x] Add link to documentation into menu

caco3:
@Slider0007 wouldn't it make sense to move the livestream menu entries into the `Data` section, like the recognition?

Slider0007:
> @Slider0007 wouldn't it make sense to move the livestream menu entries into the `Data` section, like the recognition?

I personally wouldn't see it related to data section, but if your intention is to place it there, no issue for me.

Issue #2013, opened 2023-02-11
Automatic reload of Data Graph Webpage


friedpa:
### The Feature

Would it be possible to add an automatic Webpage reload of the "Data Graph" Page, as it is implmented in tne "Overview" page? Would be great if the reload time would be adjustable.

PR #2386, opened 2023-05-02
WebUI Update

Slider0007:
Updated WebUI -> also addresses #1125, #1142, #1932, #2013, #2225, #2347, #2358

Please feel free to modify this PR for your needs or cherry-pick the parts you'd like to use.

------------------------------------
!!! Be aware of the modified wordings I used: !!!
- "Number" --> "Number sequence"
- "Alignment mark" / "reference" -> "Alignment marker"

------------------------------------
## General:
-	More mobile friendly design (no 100% mobile pages, but better than before) (#2225)
  o	Larger buttons
  o	Alignment of elements
  o	Harmonization of screens

-	Files deleted:
  o	edit_check.html
  o	test.html
  o	testcnn.html

#### _QUESTIONS / UNCLEAR / TODOs:_
- [x] Are the deleted files used for anything which is not obvious?
- [x] Updated usage description on the respective pages needs to checked / verified (correct? / consistence?)

## Index page / Overview page:
CHANGES / ADDED:
-	Values / data on left hand side
-	Image scaling down on smaller screens
-	Manual Control Menu (Only visible if HA Discovery is enabled)
  o	Resend of HA Discovery
  o	Possible to add more control functions in future (e.g. Trigger a round start)
-	Add buttons to open camera livestream (System -> Livestream)

## Configuration page:
CHANGES / ADDED:
-	Validate inputs (#2347)
  o	Prevent of input out of range
  o	Red frame / background, if out of range or empty (could happen, if value is modified in config.ini or with older versions)
-	Harmonization of visualization
  o	Hide section if not used (like it is already done for GPIO section)
  o	Larger buttons and other elements
  o	Alignment of elements
-	Implement a short usage description like it’s available in initial setup mode. I think it could be also useful to have it also in regular web interface (description section can be opened if needed)
-	Updated some static texts

FIXES:
-	Fix expert view issue (GPIO part was partly displayed even not activated)
-	Missing WebUI handling for “Flip Image Size” and “Mirror Image”

## Reference image
CHANGES / ADDED:
-	Button “Create new reference” takes an image with taking care of settings, lock this button and unlock the update image button -> Avoid confusion having two buttons which take images
-	Validate inputs
o	Prevent of input out of range
o	Red frame / background, if out of range or empty (could happen, if value is modified in config.ini or with older versions)
o	Brightness, contrast and saturation as slider input (-2 .. 2)
-	Release save button only if a new image is taken
-	Lock save button if save was done (avoid multiple saves)
-	Rename “take image” to “update image”
-	Implement a short usage description like it’s available in initial setup mode. I think it could be also useful to have it also in regular web interface (description section can be opened if needed)
-	General graphical alignment / streamlined
o	Move “save” button above image (All buttons in one place and visible without scrolling)
-	Remove all direct links because they disturb the setup logic (step by step)

FIXES:
-	Add value of rotation fine tune parameter to initial rotate parameter value (e.g. -35 + -0.2: -35.2 --> before only pre-rotation parameter was used in config, fine tune parameter was visualized but not written back to config)

## Alignment marker
CHANGES / ADDED:
-	Filename input converted to only output field
-	Release save button only if a marker update triggered
-	Lock enhance image contrast button is save button is pressed
-	Lock save button after saving (avoid multiple saves)
-	Harmonize the words “marks” and “reference” to “marker”
-	Implement a short usage description like it’s available in initial setup mode. I think it could be also useful to have it also in regular web interface (description section can be opened if needed)
-	General graphical alignment / streamlined
  o	All buttons above the image (harmonized with reference creation screen, all buttons visible without scrolling)
-	Remove all direct links because they disturb the setup logic (step by step)
-	Avoid redirect and reloading of folder content to speed up page loading (improves #2358)

#### _QUESTIONS / UNCLEAR:_
- [x] What’s the purpose of modifying the filenames of the alignment marker?

 
## Digit / Analog ROIs
CHANGES / ADDED:
-	Implement a short usage description like it’s available in initial setup mode. I think it could be also useful to have it also in regular web interface (description section can be opened if needed)
-	General graphical alignment / streamlined
  o	All buttons above the image (harmonized with the  reference creation and alignment screen, all buttons visible without scrolling)
  o	Harmonize text/button sizes and button positions
  o	Respect a max width (width of image) in design to have it more streamlined for mobile usage
-	Remove all direct links because they disturb the setup logic (step by step)
-	Rename “number” with “number sequence” to separate it more from the meaning of a single number (only “number” is ambiguous and it’s not clear that this could mean a sequence of multiple ROIs (even analog ROIs+digits ROIs), which is most likely the default use case). The renaming is only done on the web interface. No change in code.
-	Implement an output where the multiplier of the selected ROI is visualized to have an indication of the correct order of the ROIs
  o	Output only depending on the order (e.g. x100)
  o	Output pending on order + decimal shift (e.g. x100)
-	Only for analog screen: Add (CCW) to ROI name shown in image when the option “counterclockwise” is selected
FIXES:
-	Handle cancel action of all prompt popups (avoid e.g. nullstring in selection boxes)
-	Interlock al not needed buttons and inputs if no ROI is available
-	Avoid empty ROI selection if switch the number sequence (if selected a higher ROI than available ROIs in the other sequence) -> This definition is part of the description
-	Avoid empty ROI selection if switch the number sequence (if no more ROI was available in one of the sequences)
-	Fix for ROI name validation (the same name was possible multiple times)

#### _QUESTIONS / UNCLEAR:_
- [x] Number sequence and ROI name validation for invalid characters only for renaming function. Why not also for creation function?
- [x] With checked “keep distance of all ROIs” the ROIs switches positon in the image when changing the order. If not checked, only position (order) in selection box changes, the position in image remains the same.  Why is the behavior different?
 

## Info page:
CHANGES / ADDED:
-	Add memory info (REST API /heap)
-	Use XMLHttpRequest() to get the infos 

## Data page
- Add automatic reload every 300s (partly implements #2013)

## Initial setup:
CHANGES / ADDED:
- Streamlined look and feel
- Try to clear HTML cache after initial setup was used (could improve the reload behavior after setup is finished, to be proved…)
- Modified explain pages, because explanation is  now part of the subpages (reference image, aligment, ROI, config)
- Add camera livestream to setup processs as step 1 (#1932)
- Add button to abort the intial setup (jump to subpage 7)
- Add button to restart the intial setup (go back to subpage 0)
- Add WIFI RSSI value (#1125)
- Add progress bar
- Reboot buttons on all pages are removed (reboot only possible on last step) (#1142)


caco3:
### Digit / Analog ROIs
> Number sequence and ROI name validation for invalid characters only for renaming function. Why not also for creation function?

As  suggested in https://github.com/jomjol/AI-on-the-edge-device/issues/2365 we could even get rid of the naming  completely and just enumerate them.

> With checked “keep distance of all ROIs” the ROIs switches positon in the image when changing the order. If not checked, only position (order) in selection box changes, the position in image remains the same. Why is the behavior different?

It was already quite complicated to add this new feature, so I kept it as simple as possible.
It is not nice but swapping the digits is not an action to be done often.

jomjol:
I really like this UI. The follwoing comment to your question:

>  With checked “keep distance of all ROIs” the ROIs switches positon in the image when changing the order. If not checked, only position (order) in selection box changes, the position in image remains the same. Why is the behavior different?

This is because if you select the box, the position of all ROIs are calculated new.

I would like to merge this into the rolling. Any concerns?

Issue #2505, opened 2023-07-08
Rate Change no longer grows with time


dockSquadron:
### The Problem

In #1839 a line of code was added to the RateChange type logic to update the lasttime with the image time in the case of a RATE TOO HIGH error. Previously, lasttime was only updated upon success. This seems to have the effect of not allowing Rate Change to catch up over time. Is this intentional? If a single jump goes over the Rate Change limit, an error is declared, but since the Prevalue is not updated yet the meter keeps running, the delta only gets larger over time and the interval time stays fixed. This causes a Rate Too High error to become unrecoverable without manually updating a prevalue. Please evaluate if this was intended functionality.

```

            if (abs(_ratedifference) > abs(NUMBERS[j]->MaxRateValue))
            {
                WriteDataLog(j);

                NUMBERS[j]->ErrorMessageText = NUMBERS[j]->ErrorMessageText + "Rate too high - Read: " + RundeOutput(NUMBERS[j]->Value, NUMBERS[j]->Nachkomma) + " - Pre: " + RundeOutput(NUMBERS[j]->PreValue, NUMBERS[j]->Nachkomma) + " - Rate: " + RundeOutput(_ratedifference, NUMBERS[j]->Nachkomma);
                NUMBERS[j]->Value = NUMBERS[j]->PreValue;
                NUMBERS[j]->ReturnValue = "";
                NUMBERS[j]->ReturnRateValue = "";
                NUMBERS[j]->lastvalue = imagetime;

                string _zw = NUMBERS[j]->name + ": Raw: " + NUMBERS[j]->ReturnRawValue + ", Value: " + NUMBERS[j]->ReturnValue + ", Status: " + NUMBERS[j]->ErrorMessageText;
                LogFile.WriteToFile(ESP_LOG_INFO, TAG, _zw);
                LogFile.WriteToFile(ESP_LOG_ERROR, TAG, _zw);
                WriteDataLog(j);
                continue;
            }
         }
```

### Version

Development-Branch: rolling (Commit: 61efe1a)

### Logfile

```shell
[0d06h34m00s] 2023-07-07T09:35:56	<ERR>	[POSTPROC] main: Raw: 2451800, Value: , Status: Rate too high - Read: 2451800 - Pre: 2451770 - Rate: 10
[...]

PR #2511, opened 2023-07-09
revert to setting lastvalue in only success case

dockSquadron:
fixes bug #2505

### DIFF
--- a/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
+++ b/code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
@@ -757,13 +757,6 @@ bool ClassFlowPostProcessing::doFlow(string zwtime)
         NUMBERS[j]->ErrorMessageText = "";
         NUMBERS[j]->Value = -1;
 
-        /* calculate time difference BEFORE we overwrite the 'lastvalue' */
-        double difference = difftime(imagetime, NUMBERS[j]->lastvalue);      // in seconds
-
-        /* TODO:
-         * We could call `NUMBERS[j]->lastvalue = imagetime;` here and remove all other such calls further down.
-         * But we should check nothing breaks! */
-
         UpdateNachkommaDecimalShift();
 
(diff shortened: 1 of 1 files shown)


Slider0007:
Be careful: Lastvalue timestamp is also used for data logging faeture. Reverting this only back will distort data logging feature (Multiple lines with same timestamp from last successful reading whenever a "bad" reading is logged).

In my opinion two separate timestamps are necessary, one for data logging feature to ensure growing timestamps for every reading and one for your need which is only updated on succesful reading.

caco3:
@Slider0007 What do you think? Nobody else has an issue with it. And as you say, it changes the behaviour,  so I would **not** merge it.

Slider0007:
> @Slider0007 What do you think? Nobody else has an issue with it. And as you say, it changes the behaviour, so I would **not** merge it.

@caco3: In my opinion, this change is necessary to fulfill the 'self-healing' feature of `Rate too high` with the `RateChange` setting, so it is correct in principle. However, the PR would have to be extended to solve both related topics. If only this PR gets merged, the problem of data logging would be open again, which was already addressed in #1839. Therefore, I would suggest using two timestamps for the contrary purposes of 'Max Rate Check purpose' (update timepstamp only on a valid reading) and 'Data Logging purpose' (update timestamp on every cycle).

SybexX:
is fixed: https://github.com/jomjol/AI-on-the-edge-device/commit/f8eb4db171c718063812c14eb58fbdb7b1ae60d3

caco3:
> is fixed: [f8eb4db](https://github.com/jomjol/AI-on-the-edge-device/commit/f8eb4db171c718063812c14eb58fbdb7b1ae60d3)

Does that mean the issue got solved in https://github.com/jomjol/AI-on-the-edge-device/pull/3180 and we can close this PR without further work?

SybexX:
Yes, exactly and @Slider0007's concerns were also taken into account/incorporated.

PR #2651, opened 2023-10-12
added support for using TLS with MQTT

LordGuilly:
3 new entries in the config section, for setting file paths for
        - Root CA
        - Client Certificate
        - Client Private Key

If these are defined, the MQTT connection uses TLS (needed for most Cloud providers)

### DIFF
--- a/code/components/jomjol_mqtt/interface_mqtt.cpp
+++ b/code/components/jomjol_mqtt/interface_mqtt.cpp
@@ -2,6 +2,9 @@
 #include "interface_mqtt.h"
 
 #include "esp_log.h"
+#if DEBUG_DETAIL_ON
+    #include "esp_timer.h"
+#endif
 #include "connect_wlan.h"
 #include "mqtt_client.h"
 #include "ClassLogFile.h"
@@ -32,6 +35,7 @@ bool mqtt_connected = false;
 
 esp_mqtt_client_handle_t client = NULL;
 std::string uri, client_id, lwt_topic, lwt_connected, lwt_disconnected, user, password, maintopic;
+std::string caCert, clientCert, clientKey;
 int keepalive;
 bool SetRetainFlag;
 void (*callbackOnConnected)(std::string, bool) = NULL;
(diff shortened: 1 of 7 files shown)


sd-card/config/config.ini
```diff
@@ -62,6 +62,9 @@ CheckDigitIncreaseConsistency = false
 RetainMessages = false
 HomeassistantDiscovery = false
 ;MeterType = other
+;CACert = /config/certs/filepath
```
caco3, line 65: I would add default file names instead of just `filepath`.

sd-card/html/edit_config_param.html
```diff
@@ -753,6 +753,39 @@ <h4>
 			<td>$TOOLTIP_MQTT_password</td>
 		</tr>
 
+		<tr class="MQTTItem">
```
caco3, line 756: The new parameters should be marked as expert parameters, see https://github.com/jomjol/AI-on-the-edge-device/pull/2651/files#diff-a5697edb14c8de88b221309f55facd517353ada57d0e18b22b2ca04306d79317R564

sd-card/config/config.ini
LordGuilly: updated with more specfici names (yet still a bit generic)

sd-card/html/edit_config_param.html
LordGuilly: marked the 3 as expert mode

code/components/jomjol_mqtt/interface_mqtt.cpp
```diff
@@ -169,6 +173,10 @@ static esp_err_t mqtt_event_handler_cb(esp_mqtt_event_handle_t event) {
             else if (event->error_handle->connect_return_code == MQTT_CONNECTION_REFUSE_NOT_AUTHORIZED) {
                 LogFile.WriteToFile(ESP_LOG_ERROR, TAG, "Connection refused, not authorized. Check username/password (0x05)");
             }
+            else {
+                LogFile.WriteToFile(ESP_LOG_ERROR, TAG, "UNHANDLED ERROR");
```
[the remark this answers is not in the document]
LordGuilly: fixed, not it replicates the log error

LordGuilly:
updated with the comments, will update the docs in the next spare time slot :-)

caco3:
I added the parameter doc pages, please have a look and let me know if the description needs improvement/changes.
https://jomjol.github.io/AI-on-the-edge-device-docs/Parameters/#section-mqtt

It looks ok for me, but I can not test it myself.

LordGuilly:
I created the PR for the docs, sorry it took me so long - I just added some comments about how each file is used. Usually the 3 are needed, although some configurations might need some :-)

PR #2769, opened 2023-12-28
Update submodules

caco3:
- Update `esp-nn` to v1.0.2
- Update `esp32-camera` to v2.0.6
- Updated `tflite-micro-esp-examples` (The repo got renamed to `esp-tflite-micro ` and the folder structure got cleaned up).


caco3:
@jomjol @haverland @Slider0007 I suggest to delete/rename the whole project folder locally and newly set the project up according to https://github.com/jomjol/AI-on-the-edge-device/blob/rolling/code/README.md#preparations

PR #2770, opened 2023-12-28
Update platform-espressif32 to 6.5.0

caco3:
This updated the ESP IDF from ` v5.0.2` to `v5.1.2`.

### Old
```
RAM:   [==        ]  15.4% (used 50312 bytes from 327680 bytes)
Flash: [========  ]  75.9% (used 1477473 bytes from 1945600 bytes)
```
### New
```
RAM:   [==        ]  15.6% (used 51004 bytes from 327680 bytes)
Flash: [========  ]  76.9% (used 1496737 bytes from 1945600 bytes)
```

### DIFF
--- a/code/components/jomjol_fileserver_ota/server_file.cpp
+++ b/code/components/jomjol_fileserver_ota/server_file.cpp
@@ -547,7 +547,7 @@ static esp_err_t download_get_handler(httpd_req_t *req)
                 /* Get value of expected key from query string */
                 if (httpd_query_key_value(buf, "readonly", param, sizeof(param)) == ESP_OK) {
                     ESP_LOGI(TAG, "Found URL query parameter => readonly=%s", param);
-                    readonly = param && strcmp(param,"true")==0;
+                    readonly = (strcmp(param,"true") == 0);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


code/components/jomjol_fileserver_ota/server_file.cpp
```diff
@@ -547,7 +547,7 @@ static esp_err_t download_get_handler(httpd_req_t *req)
                 /* Get value of expected key from query string */
                 if (httpd_query_key_value(buf, "readonly", param, sizeof(param)) == ESP_OK) {
                     ESP_LOGI(TAG, "Found URL query parameter => readonly=%s", param);
-                    readonly = param && strcmp(param,"true")==0;
```
caco3, line 550: @jomjol Not sure what your intention here was with the old check but it would now throw this error: `the address of 'param' will never be NULL [-Werror=address]`

jomjol:
Updated the sntp part. There seems to be a change from v5.1 onwards.

jomjol:
@caco3 can you check, afterwards we can commit to rolling

jomjol (approved):
Tested changes on test device and no problems observed

caco3:
Thanks @jomol for fixing SNTP

PR #2774, opened 2023-12-28
Enhance busy notification

caco3:
Updating the markers and enhancing the contrast is only possible while no round is active.
Until now we simply showed an error and the user had to retry until the round completed.
With this MR, the UI shows an overlay message with the round status and retries automatically:
grafik

### DIFF
--- a/sd-card/html/edit_alignment.html
+++ b/sd-card/html/edit_alignment.html
@@ -67,7 +67,6 @@
             bottom: 0;
             background-color: rgba(0,0,0,0.8);
             z-index: 2;
-            cursor: pointer;
         }
 
         #overlaytext{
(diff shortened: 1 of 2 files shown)


Slider0007:
I did a quick test: Pushing the button at the wrong time (e.g. to early before page is fully loaded) results in a boot loop. As long you stay on the Alignment marker page, boot loop continues, only after switching to another page, system is booting again.

I haven't analyized it, but I assume due to the start of periodic check without device is in correct state, system crashes again and again.

image

image

```
I (15375) MAIN: Initialization completed successfully

abort() was called at PC 0x401b9ae2 on core 1


Backtrace: 0x40083a3d:0x3ffe3050 0x4008ee21:0x3ffe3070 0x40094e35:0x3ffe3090 0x401b9ae2:0x3ffe3100 0x40199d51:0x3ffe3120 0x40189c67:0x3ffe3140 0x40100d15:0x3ffe3160 0x401026e7:0x3ffe3190 0x400e6a25:0x3ffe33a0 0x400e5838:0x3ffe33e0 0x400e5f7e:0x3ffe3490 0x400e4a18:0x3ffe34b0 0x401cf8fa:0x3ffe34d0 0x400e4b5d:0x3ffe34f0
```

caco3:
@Slider0007 Thanks for spotting that. I only changed javascript code but it calls the REST API every 5 seconds once the button got pressed. The crash is due to an underlying issue we have. I will try to find it.

There are 2 issues with the old notifications:
- We can not update the text of a notification once it got created.
- We had several complains because people do not understand how it works underneath. And if you have only a short gap between the rounds it can be very tricky to catch it.

caco3:
@Slider0007 I failed to reproduce the crash. I even bombarded the REST API with the `editflow` and `statusflow` calls but it never crashed. Can you find a way to reproducible crash it?

Slider0007:
>  @Slider0007 I failed to reproduce the crash. I even bombarded the REST API with the `editflow` and `statusflow` calls but it never crashed. Can you find a way to reproducible crash it?

I opened the page and immediately pushed a button. I tested it twice just to make sure it was not a single event. Maybe only on my side and also not a realistic case. Just stressed it because it's new...

Anyway, just go ahead with your approach.

jomjol:
A good idea, I push this into rolling and see if others see this failure, too.

PR #2789, opened 2024-01-03
Show chip infos on info page

### LINKED ISSUE #2432: Show chip revision on info page
### The Feature

We should show the chip information also on the info page.
Partially it is already in the log, eg.
```
<INF> [MAIN] Device info: CPU cores: 2, Chip revision: 300
```

The USB log shows this: 
```
I (1694) cpu_start: Chip rev:         v3.0
```

See function `get_device_info()`

The 300 is build like: `chip revision number (in format MXX; where M - wafer major version, XX - wafer minor version) `

See https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/misc_system_api.html#chip-version

flooxo:
Closes #2432

This PR implements that chip information such as chip count, chip revision and chip features are displayed on the info page

### DIFF
--- a/code/main/server_main.cpp
+++ b/code/main/server_main.cpp
@@ -170,6 +170,21 @@ esp_err_t info_get_handler(httpd_req_t *req)
         httpd_resp_sendstr(req, zw.c_str());
         return ESP_OK;        
     }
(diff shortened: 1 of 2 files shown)


caco3:
Thanks!

PR #2842, opened 2024-01-31
fix(tflite): Fix memory leaks in tflite integration

Slider0007:
- Fix memory leaks in TFLite integration
  - Fix OP resolver loading
  - Fix leak in error reporter handling (missing deletion) 
    --> reporter functionality deleted entirely (optional feature and was not fully implemented at all)
- Cleanup header includes and unsued tlfite defines


friedpa:
Good Morning, I have tried your fix, but there is still a small memory leak left:

grafik

PR #2856, opened 2024-02-02
log MQTT messages to logfile instead just to UART

caco3:


caco3:
Might lead to `vfs_fat: open: no free file descriptors` due to more logging!

PR #2799, opened 2024-01-10
Support crop image on sensor, grayscale, and auto exposure level

jasaw:
Features:
1. Camera takes 1600x1200 resolution cropped to 640x480, or 800x600 cropped to 640x480 (or 320x240 depends on config). Only tested extensively on OV2640 camera.
2. Grayscale mode works with brightness/contrast/saturation control.
3. Adjustable auto exposure level, which is very important for certain lighting conditions.
4. Option to choose "Auto Exposure Control 2" algorithm for auto exposure. This might fix some under-exposed or over-exposed issues.
5. Negative photo effect might work better with existing digit models.

All these features are done by the camera itself, so no impact on memory usage or performance of the ESP32.

Discussion here: https://github.com/jomjol/AI-on-the-edge-device/discussions/2733

This is my test setup where the camera is taking 1600x1200 and cropping 640x480 out of it.
gasmeter_zoomed_negative_1

PR #2885, opened 2024-02-10
Enhanced camera functionality

caco3:
Original Pull Request: https://github.com/jomjol/AI-on-the-edge-device/pull/2799, core functionality contributed by @jasaw 

Features:
1. Camera takes 1600x1200 resolution cropped to 640x480, or 800x600 cropped to 640x480 (or 320x240 depends on config). Only tested extensively on OV2640 camera.
2. Grayscale mode works with brightness/contrast/saturation control.
3. Adjustable auto exposure level, which is very important for certain lighting conditions.
4. Option to choose "Auto Exposure Control 2" algorithm for auto exposure. This might fix some under-exposed or over-exposed issues.
5. Negative photo effect might work better with existing digit models.

**All these features are done by the camera itself, so no impact on memory usage or performance of the ESP32.**

Discussion here: https://github.com/jomjol/AI-on-the-edge-device/discussions/2733

This PR also enhances the configuration page and the Reference Image page.

This is my test setup where the camera is taking 1600x1200 and cropping 640x480 out of it.
gasmeter_zoomed_negative_1


caco3:
@jasaw I added all the new parameters to the config page.
Also I enhanced the parameter section on the reference image page.
There are no firmware code changes, nevertheless, can you please quickly test if everything still works for you? I only tested it briefly.

Just go to https://github.com/jomjol/AI-on-the-edge-device/actions?query=branch%3Afeature%2Fcrop_image and open the most top build. Then scroll down and use the AI-on-the-edge-device__update__* file for an OTA update.

jasaw:
@caco3 I found a few issues from my initial testing.

1. I get this error on serial port when I go to Settings -> Configuration. Error appears when I navigate to reference image page too. I don't know whether this error is related to this PR, so reporting it anyway.
```
E (465216) SERVER HELP: Failed to read file: /sdcard/html/img/flipImageSize.png
W (465216) httpd_txrx: httpd_resp_send_err: 404 Not Found - <pre>



        _
    .__(.)< ( oh oh! This page does not exist! )
    \___)


                You could try your <a href=index.html target=_parent>luck</a> here!</pre>
<script>document.cookie = "page=overview.html"</script>
```

2. The Sharpness range is from -4 to +3 where -4 is auto-sharpness.
sharpness_range

3. From my understanding of this new interface, the `*)` denotes a parameter that is only applied on the next Update Reference Image. If that is the case, then everything on that page requires `*)` except for
- mirror image
- flip image size
- rotation angle & fine tune
- negative
reference_image


<br>
Other than that, crop image feature appears to be working as expected.

Btw, I like the improved interface that you've done here.

caco3:
> ```
> E (465216) SERVER HELP: Failed to read file: /sdcard/html/img/flipImageSize.png
> W (465216) httpd_txrx: httpd_resp_send_err: 404 Not Found - <pre>
> ```

I saw this as well, it is already fixed in rolling

> 
>    The Sharpness range is from -4 to +3 where -4 is auto-sharpness.

Thanks, fixed


>   From my understanding of this new interface, the `*)` denotes a parameter that is only applied on the next Update Reference Image. If that is the case, then everything on that page requires `*)` except for
> 
> 
>     * mirror image
> 
>     * flip image size
> 
>     * rotation angle & fine tune
> 
>     * negative

Thanks, I missed some, fixed

> Other than that, crop image feature appears to be working as expected.
> 
> Btw, I like the improved interface that you've done here.

Thanks, I then will merge it.
Thanks for your contribution!

caco3:
@jasaw Do you know if with this PR by default (Reference image never saved, thus using the defaults of the firmware) there is a difference in the image quality?

There seem to be complains that the images are more blurry, but on my tests I do not see this change.
I only can see that there is a slight difference in the "white correction".

See https://github.com/jomjol/AI-on-the-edge-device/issues/2901

Could it be that the "sharpness" parameter has an impact?

PR #2922, opened 2024-02-16
Add waittime param to ref page + renamed HTML files

caco3:
# TODO
- [ ] create new PR but only with essential changes

# Old
- Renamed `edit_config.html` → `edit_config_raw.html` and call it "raw" edit mode instead of "direct" edit mode.
- Renamed `edit_config_param_template.html` → `edit_config_template.html`
- Added the parameter `WaitBeforeTakingPicture` to the Reference Image page.


caco3:
already implemented with different PR

PR #3081, opened 2024-05-22
add Prometheus/OpenMetrics exporter

### LINKED ISSUE #1345: Prometheus exporter
### The Feature

Hello

It would be great, if it's possible to have a prometheus exporter on /metrics url, with similar data as in MQTT messages

Regards
Laszlo

henrythasler:
Adds a new http-endpoint `/metrics` that provides metrics according to [Prometheus](https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format) (similar to [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format)) text format. Fixes #1345. The metrics mostly follow the design of @hansmi's [Prometheus Exporter](https://github.com/hansmi/prometheus-ai-on-the-edge-device-exporter/).

The main metric is probably the current value (ai_on_the_edge_device_flow_value) of the main sequence. Multiple sequences are supported via tags. This metric only provides valid readout (`no error`). Besides that, a bunch of device properties are exposed.

I'm also creating this PR to get some early feedback and advice from the community on what could be improved.

### Simple Prometheus Scrape Config

```
  - job_name: watermeter
    scrape_interval: 300s
    metrics_path: /metrics
    static_configs:
      - targets: ['192.168.178.23']
```

### Sample response

```
$ curl -v http://192.168.178.23/metrics
*   Trying 192.168.178.23:80...
* Connected to 192.168.178.23 (192.168.178.23) port 80 (#0)
> GET /metrics HTTP/1.1
> Host: 192.168.178.23
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Content-Length: 1043
< Access-Control-Allow-Origin: *
< 
# HELP ai_on_the_edge_device_flow_value current value of meter readout
# TYPE ai_on_the_edge_device_flow_value gauge
ai_on_the_edge_device_flow_value{sequence="main"} 239.2208
# HELP ai_on_the_edge_device_cpu_temperature_celsius current cpu temperature in celsius
# TYPE ai_on_the_edge_device_cpu_temperature_celsius gauge
ai_on_the_edge_device_cpu_temperature_celsius 41
# HELP ai_on_the_edge_device_rssi_dbm current WiFi signal strength in dBm
# TYPE ai_on_the_edge_device_rssi_dbm gauge
ai_on_the_edge_device_rssi_dbm -63
# HELP ai_on_the_edge_device_memory_heap_free_bytes available heap memory
# TYPE ai_on_the_edge_device_memory_heap_free_bytes gauge
ai_on_the_edge_device_memory_heap_free_bytes 714083
# HELP ai_on_the_edge_device_uptime_seconds device uptime in seconds
# TYPE ai_on_the_edge_device_uptime_seconds gauge
ai_on_the_edge_device_uptime_seconds 6760
# HELP ai_on_the_edge_device_rounds_total data aquisition rounds since device startup
# TYPE ai_on_the_edge_device_rounds_total counter
ai_on_the_edge_device_rounds_total 8
* Connection #0 to host 192.168.178.23 left intact
```
### Resource Impact

Resource | total [bytes] | current [bytes] | this PR [bytes] | change [bytes]
---|---|---|---|---
RAM | 327680 | 51040 | 51040 | 0
Flash | 1945600 | 1496605 | 1499029 | +2424


code/components/jomjol_flowcontroll/ClassFlowControll.cpp
```diff
@@ -927,3 +927,26 @@ string ClassFlowControll::getJSON()
...
+ **/
+string ClassFlowControll::getOpenMetrics(string prefix)
+{
+    string res = "";
+    std::vector<NumberPost*> *numbers = flowpostprocessing->GetNumbers();
+    
+    for (int i = 0; i < (*numbers).size(); ++i) {
+        // only valid data is reported (https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#missing-data)
+        if((*numbers)[i]->ReturnValue.length() > 0) {
+            res += prefix + "flow_value{sequence=\"" + (*numbers)[i]->name + "\"} " + (*numbers)[i]->ReturnValue + "\n";
```
hansmi, line 943: The name name may contain literal `"`, requiring escaping (something converting to JSON should be good enough).

code/components/jomjol_flowcontroll/ClassFlowControll.cpp
```diff
@@ -927,3 +927,26 @@ string ClassFlowControll::getJSON()
...
+ * @returns the string containing the text wire format of the MetricFamily
+ **/
+string ClassFlowControll::getOpenMetrics(string prefix)
+{
+    string res = "";
+    std::vector<NumberPost*> *numbers = flowpostprocessing->GetNumbers();
+    
+    for (int i = 0; i < (*numbers).size(); ++i) {
+        // only valid data is reported (https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#missing-data)
+        if((*numbers)[i]->ReturnValue.length() > 0) {
```
hansmi, line 942: `(*numbers)[i]` is used multiple times. Storing it in a variable may generate more efficient code.

code/components/jomjol_flowcontroll/MainFlowControl.cpp
```diff
@@ -474,6 +474,81 @@ esp_err_t handler_json(httpd_req_t *req)
...
+#endif
+
+    ESP_LOGD(TAG, "handler_openmetrics uri: %s", req->uri);
+
+    if (bTaskAutoFlowCreated)
+    {
+        httpd_resp_set_hdr(req, "Access-Control-Allow-Origin", "*");
+        httpd_resp_set_type(req, "text/plain"); // application/openmetrics-text is not yet supported by prometheus
+
+        const string metricNamePrefix = "ai_on_the_edge_device_";   // FIXME: move to config-file
```
hansmi, line 506: There should be no need for that FIXME. Configurable metric names just make writing rules difficult.

code/components/jomjol_flowcontroll/MainFlowControl.cpp
```diff
@@ -474,6 +474,81 @@ esp_err_t handler_json(httpd_req_t *req)
...
+        response += "# HELP " + metricNamePrefix + "rounds_total data aquisition rounds since device startup\n# TYPE " 
+            + metricNamePrefix + "rounds_total counter\n" 
+            + metricNamePrefix + "rounds_total " + std::to_string(countRounds) + "\n";
+
+        // the response always contains at least the metadata (HELP, TYPE) for the MetricFamily so no length check is needed
+        httpd_resp_send(req, response.c_str(), response.length());
+    }
+    else
+    {
+        httpd_resp_send_err(req, HTTPD_403_FORBIDDEN, "Flow not (yet) started: REST API /metrics not yet available!");
```
hansmi, line 541: HTTP 4xx is a client-side error. HTTP 5xx, in particular [HTTP 503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) seems more appropriate for this case.

code/components/jomjol_flowcontroll/MainFlowControl.cpp
```diff
@@ -474,6 +474,81 @@ esp_err_t handler_json(httpd_req_t *req)
...
+        httpd_resp_set_hdr(req, "Access-Control-Allow-Origin", "*");
+        httpd_resp_set_type(req, "text/plain"); // application/openmetrics-text is not yet supported by prometheus
+
+        const string metricNamePrefix = "ai_on_the_edge_device_";   // FIXME: move to config-file
+
+        // get current measurement (flow)
+        string response = flowctrl.getOpenMetrics(metricNamePrefix);
+
+        // CPU Temperature
+        response += "# HELP " + metricNamePrefix + "cpu_temperature_celsius current cpu temperature in celsius\n# TYPE "
```
hansmi, line 512: There are many string concatenations and repetitions. I contributed parts of [Tasmota's Prometheus endpoint](https://github.com/arendst/Tasmota/blob/v14.0.0/tasmota/tasmota_xsns_sensor/xsns_75_prometheus.ino) and found that moving such things to helper functions reduces the size of the generated code considerably (a concern on limited microcontrollers).

code/components/jomjol_flowcontroll/MainFlowControl.cpp
henrythasler: While I totally agree with you, the [enum with the response codes](https://github.com/espressif/esp-idf/blob/v5.2/components/esp_http_server/include/esp_http_server.h#L555) does not provide a 503. I guess that's the reason why all the other handler also use 403 instead.

code/components/jomjol_flowcontroll/MainFlowControl.cpp
henrythasler: I moved some of the business logic and helper functions into a separate component to improve readability. Code size was also reduced a bit.

henrythasler:
@hansmi, I have refactored the PR quite a bit according to your your suggestions. Feel free to review again.

hansmi:
@henrythasler, I'll be taking a look at the code changes as a separate review. Generally I'd suggest to break out the "fix homeassistant service discovery" changes to a separate pull request.

code/components/jomjol_flowcontroll/MainFlowControl.cpp
hansmi: I'd say `HTTPD_500_INTERNAL_SERVER_ERROR` would've been a better choice given the available codes (it's still indicative of a server problem), but if all other handlers use 403 as well your code is at least consistent with those.

code/components/openmetrics/openmetrics.cpp
```diff
@@ -0,0 +1,44 @@
+#include "openmetrics.h"
+
+using namespace std;
+
+/**
+ * create a singe metric from the given input
+ **/
+string createMetric(const string &metricName, const string &help, const string &type, const string &value)
+{
+    return "# HELP " + metricName + " " + help + "\n# TYPE " + metricName + " " + type + "\n" + metricName + " " + value + "\n";
```
hansmi, line 10: Improved formatting by adding linebreaks:

```cpp
return
  "# HELP " + metricName + " " + help + "\n" +
  "# TYPE " + metricName + " " + type + "\n" +
  metricName + " " + value + "\n";
```

code/components/openmetrics/openmetrics.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Generate the MetricFamily from all available sequences
+ * @returns the string containing the text wire format of the MetricFamily
+ **/
+string createSequenceMetrics(string prefix, std::vector<NumberPost *> *numbers)
```
hansmi, line 17: `numbers` can be a const reference: `const std::vector<NumberPost *>& numbers`. The item type could also be const, but that'd require changes to `getNumbers`'s return type.

henrythasler:
@hansmi, you are correct. I will create a new PR for the homeassistant change.

code/components/openmetrics/openmetrics.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    string res = "";
```
hansmi, line 19: The default initialization should not be necessary with `string`.

code/components/openmetrics/openmetrics.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    for (int i = 0; i < (*numbers).size(); ++i)
```
hansmi, line 21: `numbers` is a `std::vector`: `for (const auto &number : numbers) { … }`. You're already using `auto` in your code, so I assume C++11 or newer is in use.

code/components/openmetrics/openmetrics.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        auto number = (*numbers)[i];
+        // only valid data is reported (https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#missing-data)
+        if (number->ReturnValue.length() > 0)
+        {
+            auto label = number->name;
+
+            // except newline, double quote, and backslash (https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf)
+            // to keep it simple, these characters are just remove from the label
+            label = ReplaceString(label, "\\", "");
```
hansmi, line 31: If you make `label` a `string` the in-place `replaceString(std::string& s, std::string const& toReplace, std::string const& replaceWith)` can be used, saving a bit of memory.

code/components/openmetrics/openmetrics.h
```diff
@@ -0,0 +1,17 @@
...
+#include <vector>
+
+#include "ClassFlowDefineTypes.h"
+
+using namespace std;
+
+string createMetric(const string &metricName, const string &help, const string &type, const string &value);
+string createSequenceMetrics(string prefix, std::vector<NumberPost*> *numbers);
+
+#endif //OPENMETRICS_H
```
hansmi, line 17: Missing EOF newline.

code/components/openmetrics/openmetrics.h
```diff
@@ -0,0 +1,17 @@
...
+#ifndef OPENMETRICS_H
+#define OPENMETRICS_H
+
+#include <string>
+#include <fstream>
+#include <vector>
+
+#include "ClassFlowDefineTypes.h"
+
+using namespace std;
```
hansmi, line 12: I don't know about the rest of the code, but `using namespace` for any namespace in a header is a good candidate for causing problems. Just refer to the types with `std::`.

code/test/components/openmetrics/test_openmetrics.cpp
```diff
@@ -0,0 +1,56 @@
...
+                             metricName + "{sequence=\"" + number_2->name + "\"} " + number_2->ReturnValue + "\n";
+    TEST_ASSERT_EQUAL_STRING(expected2.c_str(), createSequenceMetrics(metricNamePrefix, &NUMBERS).c_str());
+}
+
+void test_openmetrics()
+{
+    test_createMetric();
+    test_ReplaceString();
+    test_createSequenceMetrics();
+}
```
hansmi, line 56: EOF newline again.

code/components/openmetrics/openmetrics.cpp
henrythasler: This will not work, because replaceString() only replaces the 1st occurrence of `toReplace`. Maybe I will implement another function to do the replacing for all occurrence.

code/main/server_main.cpp
```diff
@@ -459,7 +459,7 @@ httpd_handle_t start_webserver(void)
     config.server_port = 80;
     config.ctrl_port = 32768;
     config.max_open_sockets = 5; //20210921 --> previously 7   
-    config.max_uri_handlers = 39; // previously 24, 20220511: 35, 20221220: 37, 2023-01-02:38             
+    config.max_uri_handlers = 40; // FIXME: Why not set to a very large value in the first place?
```
caco3, line 462: Because this has an impact on the memory usage

caco3:
Please also extend the documentation at https://github.com/jomjol/AI-on-the-edge-device-docs/blob/main/docs/REST-API.md

code/main/server_main.cpp
henrythasler: There is some (IMHO: weak) evidence, that this value impacts memory usage:

`config.max_uri_handlers = 40;`|`config.max_uri_handlers = 1040;`
---|---
image|image

I would consider the ~6 Bytes per increment insignificant but we should maybe give some useful info in the comment or remove it altogether.

henrythasler:
> Please also extend the documentation at https://github.com/jomjol/AI-on-the-edge-device-docs/blob/main/docs/REST-API.md

Done: https://github.com/jomjol/AI-on-the-edge-device-docs/pull/55

code/main/server_main.cpp
caco3, line 462:
```suggestion
    config.max_uri_handlers = 40; // Make sure this fits all URI handlers. Memory usage in bytes: 6*max_uri_handlers
```

PR #3160, opened 2024-08-04
Fix building with `-D DEBUG_DETAIL_ON`

slovdahl:
Building with `-D DEBUG_DETAIL_ON` has been broken since #3029.


caco3:
What is the error message?
I think there is a better way than just remove those log messages.

SybexX:
res and quality no longer exist as they have been replaced by CCstatus.ImageFrameSize and CCstatus.ImageQuality

caco3:
Ah, ok, right

caco3:
@slovdahl  Thanks for the Pull request!

PR #3195, opened 2024-08-25
Add rate threshold parameter

SybexX:
https://github.com/jomjol/AI-on-the-edge-device/issues/3143


caco3:
I just tested it and it seems to work as expected:
```
Parameter: 5
Previous Value: 697.41240
Raw Value:      697.41237 -> -0.00003
Value Status:   no error
```
When it is disabled, I get a `Neg. Rate - Read: - Raw: 697.41237 - Pre: 697.41240 ` error (as expected).

- [ ] How ever on the config page I would suggest to move the parameter up to under the `Allow Negative Rate` and rename it to `Neg. Rate Threshold`.

Also it would be great to complete the documentation before we release it.
- [ ] To make it clearer I would suggest to mention that it is relative to the smalles ROI. Eg.
   > Maximum allowed change between two readings, **related to the smallest ROI. I**f ...

  Or does it actually relate to the extended resolution of the smallest ROI (0.00001) in my case)? Thats not clear to me.
- [ ] And the threshold is documented to be `0.05`, how ever is `2`.

SybexX:
Unbenannt

SybexX:
As it was before, this is only applied to the last number and the default value was previously 2.
0.05 is MaxRateValue and has nothing to do with it.

SybexX:
Unbenannt

caco3:
> 0.05 is MaxRateValue and has nothing to do with it.

You are right, Im sorry, I looked that up wrongly.


## Param Documentation:
As for the picture:
Could you change the font? To use Arial or another standard font? Comic font IMO is not really suitable for a documentation and is hard to read.

As for the "Notes" box:
We should remove it as long as we do not use it yet. 

In the example, please rename `Prevaluehysterese` to `Value`. Becasue people using the UI might get confused sinc ethe parameter there is called differently.
In general it would be better if we could name the parameter and its UI name the same.
Ah, and [Hysteresis](https://en.wikipedia.org/wiki/Hysteresis) is spelled differently in English :)

SybexX:
Unbenannt

caco3:
great work!

caco3:
I adjusted the param doc slightly.

According to the picture, the threshold is also used for positive values, right?
If so, calling it `Negative Rate Threshold` is misleading. Should we instead just call it `Change Threshold`?
And if so, I would move it to underneath `Maximum Rate Type` in the config page. as it is kind

Sorry for all my mind changes, you do a great job and I think it gets better and better as we think into it.

P.S: in the image, I would call it meter value instead of ounter value. i think the later is the more official term we use.

SybexX:
yes, It is used for positive and negative comparison values.
Or do we want to rename it to “ReadValueThreshold”?

SybexX:
Or "ChangeRateThreshold"?

caco3:
`ChangeRateThreshold` sounds good.
Will you also update the parameter?

SybexX:
Unbenannt

SybexX:
Unbenannt

Issue #3310, opened 2024-10-06
add contributors name in readme 


Ayushjhawar8:
### The Feature

**Title:** Request to Add Contributor Profiles to README

### Overview
I would like to propose the addition of contributor profiles to the README file of this repository. Highlighting contributors can encourage collaboration and recognize the hard work of individuals who have contributed to the project.

### Proposed Changes
- Create a section in the README file dedicated to contributors.
- Include the following information for each contributor:
  - GitHub profile link
  
### Benefits
- Acknowledges the efforts of contributors, fostering a sense of community.
- Provides visibility to contributors and encourages new developers to participate.
  
Thank you for considering this proposal!

Ayushjhawar8:
assign me if you like the idea

caco3:
Can this section be generated automatically?
Because i would not want to have to update it manually!

Feel free to provide a Pull request.

caco3:
new approach, only trigger it manually, see https://github.com/jomjol/AI-on-the-edge-device/pull/3325

Ayushjhawar8:
oh cool now i think its perfect to use

PR #3312, opened 2024-10-06
feat: #3310 Automatically Update Contributor List 

Ayushjhawar8:
#### Overview
This pull request introduces a GitHub Action workflow that automatically updates the contributor names in the README file whenever a new commit is pushed to the main branch.

#### Changes Made
- **GitHub Action Workflow**: A new workflow file named `main.yml` has been added under `.github/workflows/`.
- **Contributor List Update**: The workflow retrieves unique contributor names from the git log and appends them to the `README.md` file under a new "Contributors" section. Here is the relevant section of the code:

#### Screenshot
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,21 @@
+on:
+    push:
+        branches:
+            - main
+
+jobs:
+    contrib-readme-job:
+        runs-on: ubuntu-latest
+        name: A job to automate contrib in readme
+        permissions:
+          contents: write
+          pull-requests: write
+        steps:
+            - name: Contribute List
+              uses: akhilmhdh/contributors-readme-action@v2.3.10
+              with:
+                  image_size: 75  
+                  use_username: true
+                  columns_per_row: 10
+              env:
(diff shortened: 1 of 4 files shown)


Ayushjhawar8:
so you see the github action automatically create and update data of the table inside readme file whenever someone pushes onto `main`

caco3:
Thanks, looks nice and automatic.

Just one question:
Basically what it does is on every commit it adds another commit.
This clutters the history and also means that one has to run `git pull` after every push + action run:
image

Thus I would suggest to only run on a release. This is possible when you add it to https://github.com/jomjol/AI-on-the-edge-device/blob/main/.github/workflows/build.yaml#L274 instead to the main.yml

Also, can you add a comment above `<!-- readme: contributors -start -->` with something like `<!-- Autogenerated, do not manually update, see https://github.com/marketplace/actions/contribute-list -->`

caco3:
I adapted my change requests directly in https://github.com/jomjol/AI-on-the-edge-device/pull/3317

Thanks for your contribution!

PR #3328, opened 2024-10-09
Update FeatureRequest.md

FarukhS52:


caco3:
Thanks. 
Although that file is no longer really maintained and better should be removed.

@jomjol I think it is not really up to date anymore and I would suggest to remove it in place of enhancement issues.

PR #3330, opened 2024-10-11
Add batches to  README.md

Ranjana761:
Added badges in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,11 @@
-# AI on the Edge Device: Digitizing Your non-digital meters with an ESP32-CAM
+`	1# AI on the Edge Device: Digitizing Your non-digital meters with an ESP32-CAM
+
+[![made-with-python](https://img.shields.io/badge/Made%20with-Python-1f425f.svg)](https://www.python.org/)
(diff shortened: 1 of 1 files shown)


caco3:
Thanks fr your proposal.
How ever I have some remarks and questions about it:
1. The project is based on C++, not Python. Python is only used to train the models, but that happens in separate repos.
2. The release batch shows a non-existent version (1.1.1). Even when I load the svg directly, it shows this (hardcoded?) version:  https://img.shields.io/github/v/release/Naereen/StrapDown.js.svg
3. Also the other numbers (forks, downloads, stars) seem to be completely random!

Ranjana761:
> Thanks fr your proposal. How ever I have some remarks and questions about it:
> 
> 1. The project is based on C++, not Python. Python is only used to train the models, but that happens in separate repos.
> 2. The release batch shows a non-existent version (1.1.1). Even when I load the svg directly, it shows this (hardcoded?) version:  https://img.shields.io/github/v/release/Naereen/StrapDown.js.svg
> 3. Also the other numbers (forks, downloads, stars) seem to be completely random!

Done with changes.

caco3:
@jomjol That was a bit fast, the download counter is not correct yet.
@Ranjana761 Do you know why it is still wrong? Can we fix it somehow? It is supposed to be `>25k`!

Issue #3331, opened 2024-10-12
Home Assistant state class and unit of measure


ykri:
### Issue

Since last HA update 2024.10 I am getting two repair issues for raw value. 
```

The entity no longer has a state class

We have generated statistics for 'ElectrMeter Raw Value' (sensor.electrmeter_raw) in the past, but it no longer has a state class, therefore, we cannot track long term statistics for it anymore.

Statistics cannot be generated until this entity has a supported state class.
If the state class was previously provided by an integration, this might be a bug. Please report an issue.
If you previously set the state class yourself, please correct it. [The different state classes and when to use which can be found in the developer documentation.](https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics)
If the state class has permanently been removed, you may want to delete the long term statistics of it from your database.
Do you want to permanently delete the long term statistics of sensor.electrmeter_raw from your database?

```
 and

```

The unit has changed

The unit of 'ElectrMeter Raw Value' (sensor.electrmeter_raw) changed to '' which can't be converted to the previously stored unit, 'kWh'.
If the historic statistic values have a wrong unit, you can update the units of the old values. The values will not be updated.
Otherwise you can choose to delete all historic statistic values, and start over

```

caco3:
I updated the HA Discover topics in https://github.com/jomjol/AI-on-the-edge-device/pull/3332
This should fix the first issue. I think the 2nd issue is just a follow-up error due to the first error.

I can't test it myself since I don't see this error. So I need you to test and confirm it is ok now!
Please test the latest (most top) build from https://github.com/jomjol/AI-on-the-edge-device/actions?query=branch%3Asupport-homeassistant-2024.10

ykri:
Still the same. See the HA MQTT debug info for raw

![image](https://github.com/user-attachments/assets/dafcc3e9-0f32-45e9-b511-7f448bb8b1e7)

caco3:
I can not reproduce this!
For me it looks ok now:
![grafik](https://github.com/user-attachments/assets/21e4a3b1-ad41-43cb-b836-281d2298aa6c)

Maybe try to delete the discovery topics and let the device re-send them.

ykri:
I tried to delete the discovery topics using MQTT Explorer but I am back to zero. Is there any other clever way to do this discovery topics deletion?

caco3:
[...]

PR #3332, opened 2024-10-12
Update the Homeasistant Discovery topics

caco3:
This PR tries to fix https://github.com/jomjol/AI-on-the-edge-device/issues/3331 and https://github.com/jomjol/AI-on-the-edge-device/issues/3333

---

Changes:
    - `raw` has now set the `State Class` to `measurement`. Before it was always set to `""`. 
    - `value` has now only set the `State Class` to `total_increasing` if the parameter `Allow Negative Rates` is **not** set. Else it uses `measurement` since the rate could also be negative. Before it was always set to `total_increasing`.
    - The `rate_per_time_unit` topic of an **Energy** meter needs a `Device Class`=`power`. For `gas` and `water` it should be `volume_flow_rate`. Before it was always set to `""`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,8 @@ Only changes since RC4 are listed:
     - `value` has now only set the `State Class` to `total_increasing` if `Allow Negative Rates` is NOT set. Else it uses `measurement` as it could go down. Before it was always set to `total_increasing`.
     - The `rate_per_time_unit` topic of an **Energy** meter needs a `Device Class`=`power`. For `gas` and `water` it should be `volume_flow_rate`. Before the `Device Class` was always set to `""`.
 
+  **:warning: Please check your Homeassistant instance to make sure it is handled correctly!**
+
 #### Bug Fixes
 Only changes since RC3 are listed:
  - None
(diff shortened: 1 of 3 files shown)


caco3:
@ykri I updated the `Device Class` according your suggestions in https://github.com/jomjol/AI-on-the-edge-device/pull/3332

Can you please test and see if it is ok now?
Please test the latest (most top) build from https://github.com/jomjol/AI-on-the-edge-device/actions?query=branch%3Asupport-homeassistant-2024.10

Frans1960:
I have an older meter and PV panels, so the meter runs backwards when the sun shines. Only Home assistant needs the State Class to **total_increasing** even when **Allow Negative Rates** is set. Can you make a fix or solution for this? Hanks

- value has now only set the State Class to total_increasing if the parameter Allow Negative Rates is not set. Else it uses measurement since the rate could also be negative. Before it was always set to total_increasing.

caco3:
> I have an older meter and PV panels, so the meter runs backwards when the sun shines. Only Home assistant needs the State Class to **total_increasing** even when **Allow Negative Rates** is set. Can you make a fix or solution for this? Hanks
> 
>     * value has now only set the State Class to total_increasing if the parameter Allow Negative Rates is not set. Else it uses measurement since the rate could also be negative. Before it was always set to total_increasing.

Please create a separate discussion for this.

PR #3358, opened 2024-10-26
feat: auto-comment workflow

Ranjana761:
## Summary 
Added a workflow which will automatically comment on newly opened PR and Issues.


caco3:
Personally I don't think we need this.
The auto response on a new issue gives the impression as if somebody of the team gets assigned to work on it. How ever we are just a few guys who do this project in our free time. None of us is able to give a hard time commitment.

Issue #3302, opened 2024-10-05
Publish GPIO settings with MQTT is not included in HA auto discovery


Dag0d:
### The Problem

When GPIO Pins are configured they are published to MQTT if "GPIO1 Enable MQTT" is checked this is as expected.
But it is not included in the HA auto discovery message.
Maybe there was a reason behind this decision, if so I´m happy to hear it. 

### Version

1.5.7 and 1.6.0 RC2

### Logfile

```shell
I guess no need for a log file
```

caco3:
The reason simply is that I was not aware of it when I implemented the discovery topics.
If somebody wants to extend it, I am happy to test it, but don't currently have the time to extend it myself.

caco3:
@Dag0d I am happy to extend the discovery topics to reflect the control topics we have.
Currently there are 3 of them:
-  `ctrl/flow_start`
-  `ctrl/set_prevalue`
- `GPIO/GPIO{PinNumber}`

They are already documented in https://jomjol.github.io/AI-on-the-edge-device-docs/MQTT-API/#control

How ever I struggle to get them working (make them show up in HA). I tried it like in https://community.home-assistant.io/t/add-entity-to-mqtt-device/540949/7 but nothing showed up in HA.
So if somebody is willing to provide the needed config topics, I will add it. if not, I will move it aside for now.

Slider0007:
@caco3: I only have it implemented for flow start:
```
.field = "flow_start",
.friendlyName = "Manual Cycle Start",
.icon = "timer-play-outline",
.deviceClass = "update",
.entityCategory = "config"

topicFull = "homeassistant/button/" + node_id + "/" + configTopic + "/config";
```

In HA:
![image](https://github.com/user-attachments/assets/df948250-e46e-4795-a777-872b9e13ce86)

caco3:
@Slider0007 Nice. Can you prepare a branch with it? I then can check to extend it.

Other question: 
Do you know why we need a payload here?
https://github.com/jomjol/AI-on-the-edge-device/blob/43f15fcba3416a53bea428bb6e06b236aa72b3a4/code/components/jomjol_mqtt/interface_mqtt.cpp#L391

Slider0007:
> @Slider0007 Nice. Can you prepare a branch with it? I then can check to extend it.
> 
> Other question: Do you know why we need a payload here?
> 
[...]

PR #3415, opened 2024-12-01
Extend Homeassistant discovery with button for flow start

caco3:
See https://github.com/jomjol/AI-on-the-edge-device/issues/3302

- [x]     ctrl/flow_start
- [x] Remove `if (_data_len > 0) { `, see https://github.com/jomjol/AI-on-the-edge-device/issues/3302#issuecomment-2422087912 

##  ctrl/flow_start
Adds a button to to Homeassistant to trigger the flow:
grafik

## out of scope for this Pull request:
- [ ]     ctrl/set_prevalue
- [ ]     GPIO/GPIO{PinNumber} -> but only those pins which have MQTT enabled

### DIFF
--- a/code/components/jomjol_mqtt/interface_mqtt.cpp
+++ b/code/components/jomjol_mqtt/interface_mqtt.cpp
@@ -388,13 +388,7 @@ bool mqtt_handler_flow_start(std::string _topic, char* _data, int _data_len)
 {
     ESP_LOGD(TAG, "Handler called: topic %s, data %.*s", _topic.c_str(), _data_len, _data);
 
-    if (_data_len > 0) {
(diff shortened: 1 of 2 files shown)


caco3:
@Slider0007 I tested your code for the flow control start and it seems to work well.
I looked into how to do it with the `set_prevalue`, how ever the problem seems to be that we can't pre-configure HA to show a useful input. I tried it with a `homeassistant/number/` (see https://github.com/jomjol/AI-on-the-edge-device/pull/3416), how ever it stays disabled and the shown slider is not useful at all:
grafik

Therefore I would keep the  `set_prevalue` out of scope.

The GPIOs should be feasible, but I haven't looked into it yet. Ideally on the HA UI they would represent the real state.

PR #3501, opened 2025-01-16
add camera deep sleep for ov5640

SybexX:
Unfortunately only available on OV3660 and OV5640.

This reduces power consumption and the ESP32 temperature (in my tests approx. 5 to 10 °C less).


SybexX:
@jasaw can you please test it with your OV5640?

jasaw:
@SybexX I've tested this on my OV5640 and it's working perfectly. The camera temperature dropped 10 degrees C. Very impressive !

PR #3504, opened 2025-01-17
feature: support manual and auto focus on OV5640

jasaw:
This is still work in progress, but I want to create a PR to get early feedback.


jasaw:
@SybexX Can I please get your thoughts on whether this is going to play nicely with your camera deep sleep changes?

SybexX:
can this also be implemented as a function? e.g. like this:

````
// HACK: testing only
CCstatus.CameraFocusEnabled = true;
CCstatus.CameraManualFocus = true;
// CCstatus.CameraManualFocusLevel = 0x01e4;
CCstatus.CameraManualFocusLevel = 0x0014;

esp_err_t CCamera::SetCamAF(void)
{
    if (CCstatus.CameraFocusEnabled && CCstatus.CamSensor_id == OV5640_PID && CCstatus.CameraAFInitSuccessful) {
        ESP_LOGI(TAG, "OV5640 and AF inited");
        sensor_t *s = esp_camera_sensor_get();
		
        if (s != NULL) {
            if (CCstatus.CameraManualFocus) {
                int rc = ov5640_manual_focus_set(s, CCstatus.CameraManualFocusLevel);
				
                if (rc == 0) {
                    ESP_LOGI(TAG, "Set manual focus level success");
                } 
                else {
                    ESP_LOGI(TAG, "Set manual focus level failed: %d", rc);
                    return ESP_FAIL;
                }
            } 
            else {
                int rc = ov5640_autofocus_set_mode(s, AF_TRIG_SINGLE_AUTO_FOCUS);
				
                if (rc == 0) {
                    ESP_LOGI(TAG, "Set single autofocus mode success");
                } 
                else {
                    ESP_LOGI(TAG, "Set single autofocus mode failed: %d", rc);
                    return ESP_FAIL;
                }
				
                ESP_LOGI(TAG, "adjusting focus");
				
                for (int i = 0; i < 10; i++) {
                    camera_fb_t *fb = esp_camera_fb_get();
                    uint8_t S_Zone[5];
                    uint8_t focus_status = ov5640_autofocus_get_status(s, S_Zone, 5);
					
                    for (int i = 0; i < 5; i++) {
                        ESP_LOGI(TAG, "Zone[%d]: 0x%02x", i, S_Zone[i]);
                    }
					
                    esp_camera_fb_return(fb);
					
                    if (focus_status == FW_STATUS_S_FOCUSING) {
                        ESP_LOGI(TAG, "Focusing: 0x%02x", focus_status);
                    } 
                    else if (focus_status == FW_STATUS_S_FOCUSED) {
                        ESP_LOGI(TAG, "Focused: 0x%02x", focus_status);
                        break;
                    } 
                    else {
                        ESP_LOGI(TAG, "Focus status 0x%02x", focus_status);
                    }
                }
            }
        }
    }
    return ESP_OK;
}

esp_err_t CCamera::ReleaseCamAF(void)
{
    if (CCstatus.CameraFocusEnabled && CCstatus.CamSensor_id == OV5640_PID && CCstatus.CameraAFInitSuccessful) {
        ESP_LOGI(TAG, "OV5640 release AF");
        sensor_t *s = esp_camera_sensor_get();
		
        if (s != NULL) {
            if (CCstatus.CameraManualFocus) {
                int rc = ov5640_manual_focus_release(s);
				
                if (rc == 0) {
                    ESP_LOGI(TAG, "Release manual focus success");
                } 
                else {
                    ESP_LOGI(TAG, "Release manual focus failed: %d", rc);
                    return ESP_FAIL;
                }
            } 
            else {
                int rc = ov5640_autofocus_release(s);
				
                if (rc == 0) {
                    ESP_LOGI(TAG, "Release autofocus success");
                } 
                else {
                    ESP_LOGI(TAG, "Release autofocus failed: %d", rc);
                    return ESP_FAIL;
                }
            }
        }
    }
    return ESP_OK;
}
````

Do you always have to set the focus before taking the picture, or can you only do this once when starting the ESP32 and it is retained?

SybexX:
Theoretically, you would have to determine the focus value when creating a reference image and then save it to ensure that the alignment and evaluation works without any problems.
Reference image with autofocus >>> Save focus value >>> Image capture with manual focus (with the values ​​determined from autofocus)

SybexX:
Here I have combined deep sleep with your auto focus:
https://github.com/SybexX/AI-on-the-edge-device/tree/test
https://github.com/SybexX/AI-on-the-edge-device/actions/runs/12836438844/artifacts/2449146461

jasaw:
@SybexX 

>Do you always have to set the focus before taking the picture, or can you only do this once when starting the ESP32 and it is retained?

For our use case, I do not recommend retaining the focus constantly. Technically, we could set the focus once and retain the focus for subsequent image captures, but the focus VCM circuit will constantly draw 100mA, which will heat up the camera module. The way the focus is adjusted is by driving the VCM coil to generate magnetic field to repel against a permanent magnet, therefore moving the focus lens. If we turn off the power to the VCM coil, the lens snaps back to the resting position.

Long story short, yes, we should set the focus on each capture and release focus after the capture. This applies to both manual and auto focus.

>Theoretically, you would have to determine the focus value when creating a reference image and then save it to ensure that the alignment and evaluation works without any problems.
Reference image with autofocus >>> Save focus value >>> Image capture with manual focus (with the values ​​determined from autofocus)

Yes, pretty much with the option of manually adjusting the focus level determined by autofocus. The autofocus sometimes gets it wrong, so may require a bit of manual adjustment.

SybexX:
hast du die angepasste firmware schon getestet?

jasaw:
@SybexX Thanks for integrating the focus stuff into your deep sleep code.

I'll rebase my branch to your `add-CameraDeepSleep-for-OV5640` branch and keep working from there.

>hast du die angepasste firmware schon getestet?

Sorry, I haven't tested your test branch where you included the focus code into your deep sleep support. I'll test it as soon as possible and report back.

SybexX:
My test fork is only intended as an example to show what I meant by “Can this also be implemented as a function?”
I have now deleted my deep sleep PR, if you want you can add it in your PR.

This is the main code for deep sleep + auto focus:
````
// only available on OV3660 and OV5640
// https://github.com/espressif/esp32-camera/issues/672
int CCamera::SetCamDeepSleep(bool enable)
{
    int ret = 0;

    if (CCstatus.CameraDeepSleepEnable != enable)
    {
        CCstatus.CameraDeepSleepEnable = enable;

        // release focus from deep sleep
        if ((CCstatus.CameraDeepSleepEnable) && (CCstatus.CamSensor_id != OV2640_PID))
        {
            ret |= ReleaseCamAutoFocus();
        }

        if (CCstatus.CamSensor_id == OV2640_PID)
        {
            // OV2640 (Normal mode >>> Standby mode = OK), (Standby mode >>> Normal mode = n.OK)
            // ret |= s->set_reg(s, 0x109, 0x10, enable ? 0x10 : 0);
            // LogFile.WriteToFile(ESP_LOG_INFO, TAG, "DeepSleep is not supported by OV2640");
        }
        else
        {
            sensor_t *s = esp_camera_sensor_get();

            if (s != NULL)
            {
                ret |= s->set_reg(s, 0x3008, 0x42, enable ? 0x42 : 0x02);
            }
            else
            {
                ret = -1;
            }
        }

        vTaskDelay(100 / portTICK_PERIOD_MS);

        // after the camera has been brought out of deep sleep, set the focus
        if ((!CCstatus.CameraDeepSleepEnable) && (CCstatus.CamSensor_id != OV2640_PID))
        {
            ret |= SetCamAutoFocus();
        }
    }

    return ret;
}

int CCamera::SetCamAutoFocus(void)
{
    int ret = 0;

    if (CCstatus.CamSensor_id == OV5640_PID)
    {
        if (CCstatus.CameraFocusEnabled && CCstatus.CameraAFInitSuccessful)
        {
            ESP_LOGI(TAG, "OV5640 and AF inited");
			
            if (CCstatus.CameraDeepSleepEnable == true)
            {
                ret |= SetCamDeepSleep(false);
                vTaskDelay(100 / portTICK_PERIOD_MS);
            }
			
            sensor_t *s = esp_camera_sensor_get();

            if (s != NULL)
            {
                if (CCstatus.CameraManualFocus)
                {
                    ret |= ov5640_manual_focus_set(s, CCstatus.CameraManualFocusLevel);

                    if (ret == 0)
                    {
                        ESP_LOGI(TAG, "Set manual focus level success");
                    }
                    else
                    {
                        ESP_LOGI(TAG, "Set manual focus level failed: %d", ret);
                    }
                }
                else
                {
                    ret |= ov5640_autofocus_set_mode(s, AF_TRIG_SINGLE_AUTO_FOCUS);

                    if (ret == 0)
                    {
                        ESP_LOGI(TAG, "Set single autofocus mode success");
                    }
                    else
                    {
                        ESP_LOGI(TAG, "Set single autofocus mode failed: %d", ret);
                        return ret;
                    }

                    ESP_LOGI(TAG, "adjusting focus");

                    for (int i = 0;  i < 10;  i++)
                    {
                        camera_fb_t *fb = esp_camera_fb_get();
                        uint8_t S_Zone[5];
                        uint8_t focus_status = ov5640_autofocus_get_status(s, S_Zone, 5);

                        for (int j = 0;  j < 5;  j++)
                        {
                            ESP_LOGI(TAG, "Zone[%d]: 0x%02x", j, S_Zone[i]);
                        }

                        esp_camera_fb_return(fb);

                        if (focus_status == FW_STATUS_S_FOCUSING)
                        {
                            ESP_LOGI(TAG, "Focusing: 0x%02x", focus_status);
                        }
                        else if (focus_status == FW_STATUS_S_FOCUSED)
                        {
                            ESP_LOGI(TAG, "Focused: 0x%02x", focus_status);
                            break;
                        }
                        else
                        {
                            ESP_LOGI(TAG, "Focus status 0x%02x", focus_status);
                        }
                    }
                }
            }
            else
            {
                ret = -1;
            }
        }
    }

    return ret;
}

int CCamera::ReleaseCamAutoFocus(void)
{
    int ret = 0;

    if (CCstatus.CamSensor_id == OV5640_PID)
    {
        if (CCstatus.CameraFocusEnabled && CCstatus.CameraAFInitSuccessful)
        {
            ESP_LOGI(TAG, "OV5640 release AF");
			
            if (CCstatus.CameraDeepSleepEnable == true)
            {
                ret |= SetCamDeepSleep(false);
                vTaskDelay(100 / portTICK_PERIOD_MS);
            }
			
            sensor_t *s = esp_camera_sensor_get();

            if (s != NULL)
            {
                if (CCstatus.CameraManualFocus)
                {
                    ret |= ov5640_manual_focus_release(s);

                    if (ret == 0)
                    {
                        ESP_LOGI(TAG, "Release manual focus success");
                    }
                    else
                    {
                        ESP_LOGI(TAG, "Release manual focus failed: %d", ret);
                    }
                }
                else
                {
                    ret |= ov5640_autofocus_release(s);

                    if (ret == 0)
                    {
                        ESP_LOGI(TAG, "Release autofocus success");
                    }
                    else
                    {
                        ESP_LOGI(TAG, "Release autofocus failed: %d", ret);
                    }
                }
            }
            else
            {
                ret = -1;
            }
        }
    }

    return ret;
}
````

jasaw:
@SybexX I've added focus control to the "create new reference" page and configuration page. I've tested it and it's working as expected.

Any reason why you closed your deep sleep PR? Did you discover problems with it? I'm thinking of adding deep sleep to this PR based on your deep sleep PR.

SybexX:
@jasaw I made some adjustments that should solve the first problem:
[AI-on-the-edge-device-feature-focus_support.zip](https://github.com/user-attachments/files/18527716/AI-on-the-edge-device-feature-focus_support.zip)

jasaw:
@SybexX Thanks for your code adjustment. Both issues that I mentioned are solved now.

SybexX:
@jasaw AI-on-the-edge-device/sd-card/html/edit_reference.html >>> Line 959 >>> WriteParameter(param, category, "TakeImage", "CamFocusManualLevel", false, true); causes an error.
You have to change it to WriteParameter(param, category, "TakeImage", "CamFocusManualLevel", false);

SybexX:
@jasaw What do you think about adding a debug output during deep sleep so that you can follow the functionality better?
````
        if (CCstatus.CamSensor_id == OV2640_PID)
        {
            // OV2640 (Normal mode >>> Standby mode = OK), (Standby mode >>> Normal mode = n.OK)
            // ret |= s->set_reg(s, 0x109, 0x10, enable ? 0x10 : 0);
            // LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "DeepSleep is not supported by OV2640");
            ESP_LOGD(TAG, "DeepSleep is not supported by OV2640");
        }
        else
        {
            sensor_t *s = esp_camera_sensor_get();
            s->set_reg(s, 0x3008, 0x42, enable ? 0x42 : 0x02);
            // LogFile.WriteToFile(ESP_LOG_DEBUG, TAG, "DeepSleep: %d", enable);
            ESP_LOGD(TAG, "DeepSleep: %d", enable);
        }
````

SybexX:
@jasaw I have prepared something, maybe you would like to include it in your PR.
https://github.com/SybexX/AI-on-the-edge-device/commit/b2f89248fb5e6a6b070e3fa002232329179de67e
https://github.com/SybexX/AI-on-the-edge-device/commit/cb39551647b831ab2af9afd6a7757e58459e06be
https://github.com/SybexX/AI-on-the-edge-device/commit/675a8cb7ee6fed8a41c5cf0e2f6c69e11aab2e5e

then you can e.g. add something like this:
````
if ((camId != "OV5640") && (camFocus == "unavailable")){
    EnDisableItem(false, param, category, "TakeImage", "CamFocusAuto", false);
    EnDisableItem(false, param, category, "TakeImage", "CamFocusManualLevel", false);
    // or hide it completely with setVisible(className, visible)
    ........................................
}
````

jasaw:
@SybexX Thanks. I've added hiding of focus control when camera does not support focus control.

SybexX:
@jasaw Something does not fit with the initialization of the focus, although my camera does not have one, it is successfully initialized and also displayed on the reference page ^^ When I try it with an OV2640, it successfully hides on the reference page.

jasaw:
@SybexX Unfortunately, there is no way to identify whether the focus VCM coil is physically attached or not. The only difference between autofocus and non-autofocus OV5640 is the presence of external VCM coil and there is no way for software to work that out.

jasaw:
@SybexX I have added code to reconfigure the camera after waking up on one of my OV5640 without autofocus, but the one with autofocus doesn't need reconfiguration. I suspect there is back powering from the autofocus circuit to the camera.

jasaw:
@SybexX The deep sleep functionality appears to be working nicely with focus control so far. I've been able to achieve 10 degrees C temperature drop !

Currently, autofocus doesn't work with the zoom mode. I suspect it confuses the autofocus algorithm. I'm going to experiment with a few things to see whether I can workaround the issue. I'll keep you updated. Manual focus has no issue at all.

SybexX:
@jasaw I think the autofocus resets the camera settings, so they probably have to be set afterwards.

change from
````
    CheckCamSettingsChanged(&_focusEnabled, &_manualFocus, &_manualFocusLevel);
    SetCamFocus(_focusEnabled, _manualFocus, _manualFocusLevel);
````
to
````
    SetCamFocus(_focusEnabled, _manualFocus, _manualFocusLevel);
    CheckCamSettingsChanged(&_focusEnabled, &_manualFocus, &_manualFocusLevel);
````

jasaw:
@SybexX Autofocus microcode constantly overwrites the zoom registers, so I have to run autofocus without zoom to find the focus level, then switch to manual focus with zoom.

I have refactored the code a little bit to change when camera config is loaded to the camera registers. I think it's cleaner to have one location that decides which config to load to the camera registers, which works well with the camera deep sleep. I also fixed a few bugs along the way. Please have a look and let me know if I have broken something.

jasaw:
@SybexX Just a quick update on my progress. After my refactor, my OV5640 without autofocus has occasional messed up camera configuration (the photo looks a bit off, like too bright or too dark as though it lost the brightness config). Weird thing is, it is working perfectly on my OV564 with autofocus. I initially thought it's caused by higher ambient temperature on my OV5640 without autofocus, but now I'm thinking the autofocus circuit has some back powering that prevents the registers from resetting during deep sleep. I'm still investigating the root cause of the issue.

SybexX:
@jasaw I also have the same problems with my OV5640, I assume that the camera hangs up or that the configurations are not transmitted properly. As far as I know, the settings are not lost in the Deepsleep mode and do not have to be set up constantly.
I once combined your new idea (camera_config.h) with deepepsleep:
https://github.com/SybexX/AI-on-the-edge-device/tree/test_cam

jasaw:
@SybexX my latest commit appears to resolve the occasional photo brightness problem for me. I'll continue to monitor it over the next few days.

I'll look into incorporating your changes after I have confirmed that the intermittent brightness has been resolved.

SybexX:
My camera still has problems with it.
If I, for example. Zoom Size: 59 Set, only that comes:
Unbenannt

SybexX:
I played something around with the zoom and unfortunately it didn't take long for the ESP32 to land in a boot looop.

jasaw:
@SybexX I've been testing it and monitoring it. I haven't been able to track down the root cause of the problem.
This is a summary of what I've observed:
- There is a difference in behaviour between OV5640 with and without autofocus. The brightness problem is more easily reproducible on the OV5640 without autofocus.
- When I write camera config to the camera before each image capture, there is a chance of messing up the captured image regardless of whether camera is put into deepsleep. That means switching between CCstatus and CFstatus quickly causes problem.
- On my test branch, I added mutex around code that touches camera config, but that made no difference, so it's not caused by concurrency.

I'm not sure what could be causing the issue. I'll continue to investigate.

SybexX:
@jasaw By testing, I found that even too often reading the camera settings causes errors in the transmission.
So you have to see how you can reduce reading and writing the camera settings to the minimum, and you may have to add a plausibility test when reading the camera settings.
In addition, if you use zoom, the camera sporadically causes errors when the quality is set below 10.
Without zoom, even a value of 6 does not cause any problems with my camera.

jasaw:
@SybexX Not sure if you have seen the discussion here. We may be facing the same issue? https://github.com/espressif/esp32-camera/issues/244

jasaw:
@SybexX 
>if you use zoom, the camera sporadically causes errors when the quality is set below 10.

From my experience with ov5640, the stable value for jpeg quality depends on the complexity of the captured image.
For example, I can get away with quality value 8 if I do grey scale and denoise value 8 because the image is smoother, so resulting jpeg is small. If I reduce denoise value to 3, or enable colour, it starts failing depending on lighting conditions (regardless of zoom).

Again, from my limited experience, jpeg quality 16 appears to work in all configurations that I've tested (colour, denoise, etc). If I do grey scale, it's safe to run with jpeg quality 12.

jasaw:
@SybexX This is just my observation, further investigation is needed.

It looks like hitting the `Overview` button on the html page repeatedly during "Taking image" step has a higher chance of causing the captured image to be corrupted. I ran this test with wifi running on core 1, everything else on core 0. Wifi is on the default min low power mode with -67dB signal strength. I can't reproduce this on my other ESP32cam with -38dB wifi signal strength.

jasaw:
@SybexX Another theory is after waking up from deepsleep, the camera needs a few frames to adjust the auto exposure and auto white balance.
See https://github.com/espressif/esp32-camera/issues/314

SybexX:
I don't know if that brings something, because this problem should actually eliminate "WaitBeforePicture".

jasaw:
@SybexX I have moved the wifi back to core 0 because the camera task is running on core 1 and we want to minimize interruption to the camera SCCB running on I2C1 interface.

You are right that moving `WaitBeforePicture` to after `PrecaptureCamSetup` should do the trick, but the `LightOnOff` has to be done before `PrecaptureCamSetup` because it needs the light for autofocus to work.
I think `WaitBeforePicture` defaults to 5 seconds but now I'm thinking whether that is necessary. I feel that calling `esp_camera_fb_get()` in a loop of 5 or 10 is more appropriate. The ov5640 auto-exposure control has auto banding where it needs a few frames to determine the best exposure and gain. I'm currently testing looping `esp_camera_fb_get()` 5 times before saving the image instead of 5s delay. From my initial testing, I haven't encountered a single badly exposed with bad white balance image yet. Only time will tell whether the bad exposure bad white balance problem is completely gone, so I'll continue to monitor it.

Regarding the corrupted image that I get occasionally, it happens regardless of `WaitBeforePicture` 5s delay and/or looping `esp_camera_fb_get()` 5 times, but I only ever see this on one of my ov5640 cameras that's running at higher ambient temperature (under direct sunlight). Now I'm suspecting that it is temperature related.
The ambient temperature here is around 40°C today and I happen to notice more corrupted image than usual. I'm currently testing my theory by running the CPU at 80MHz to keep it as cool as possible and I've reduced the `xclk_freq_hz` from 20MHz down to 15MHz.

I've also increased the minimum quality value for ov5640 to 16 just to make sure that I avoid the corrupted jpeg issue when quality value is set too low.

All these 3 changes on still on my local branch. I'll continue to monitor and report back as usual.

ps: I found a bug in the CPU frequency config handling code.

jasaw:
@SybexX A quick update:

On my local branch, I have re-incorporated the 5 seconds `WaitBeforePicture` before looping `esp_camera_fb_get()` 5 times. I'm not sure whether I need both, but that can be optimized later.

I have also reduced the `xclk_freq_hz` down to 10MHz and this made a HUGE difference to the OV5640 image quality. Compared to 20MHz clock, the 10MHz gives me less noise, sharper image, well balanced exposure. I have auto-denoise and auto-exposure enabled. I have observed a change in image quality when I play around with the clock frequency and have seen reports of this on forums too. The downside of reducing the clock frequency is the reduced frame rate, but we do not need high frame rate anyway.

Slider0007:
> I have also reduced the `xclk_freq_hz` down to 10MHz and this made a HUGE difference to the OV5640 image quality. Compared to 20MHz clock, the 10MHz gives me less noise, sharper image, well balanced exposure. I have auto-denoise and auto-exposure enabled. I have observed a change in image quality when I play around with the clock frequency and have seen reports of this on forums too. The downside of reducing the clock frequency is the reduced frame rate, but we do not need high frame rate anyway.

@jasaw: Interesting :-)
I also have my OV5640 camera running with 10Mhz only for board responiveness and lower camera temperatures reason for a longer time without issues. In my environment I couldn't see that much difference between 20Mhz an d 10Mhz in terms of image quality, though.
I also have OV2640 which has horribly image quality and transmission issues using 8Mhz, but running without issues and decent image quality at 10Mhz. It seems it depends somehow on camera module quality, board/camera module combination and maybe environment.
Haven't tested with auto focus variant nor sleep mode...

jasaw:
Now I highly suspect some OV5640 camera modules are more susceptible to electrical noise than others and therefore some are not stable at 20MHz xclk. This is probably made worse by noisy power supply. Best workaround that I've found is to run it at 10 MHz. I've been testing both my OV5640 at 10MHz for 1 week now and I haven't seen a single corrupted image. @SybexX @Slider0007 Have you guys seen any corrupted image running at 10 MHz?

The better image quality when running at 10 MHz is a mystery but I'll happy accept better quality image with lower camera temperature. That's win-win for me :-)

The other problem that I was observing on one of my ov5640 was caused by high temperature. On extremely hot days, under the direct sunlight, my ov5640 gets overheated and I ended up the web interface saying "Flow not yet created", so that's my hardware fault for running it too hot. I better heat management on my hardware.

I believe the occasional corrupted image mystery is solved now. I'll continue to monitor it and do a full test with my ov5640+AF camera module.

Slider0007:
> Now I highly suspect some OV5640 camera modules are more susceptible to electrical noise than others and therefore some are not stable at 20MHz xclk. Best workaround that I've found is to run it at 10 MHz. I've been testing both my OV5640 at 10MHz for 1 week now and I haven't seen a single corrupted image. @SybexX @Slider0007 Have you guys seen any corrupted image running at 10 MHz?

I haven't any issues so far while operating OV5640 with 10Mhz. All my devices (not all with OV5640, though) are operated either with 8Mhz or 10Mhz. Going lower than 8Mhz, the image quality is decreasing somehow (artifacts, disturbing lines, etc.). Not every camera module behaves the same in this respect either, regardless of whether it's an OV2640 or OV5640.

> The better image quality when running at 10 MHz is a mystery but I'll happily accept better quality image with lower camera temperature. That's win-win for me :-)

Fully agree.

jasaw:
@SybexX I've been testing without the 5s `WaitBeforePicture` delay and these are my findings:
- `WaitBeforePicture` is not needed as far as I can tell. I am looping image capture 10 times before the actual capture and the image quality has been consistently good.
- The 5s `WaitBeforePicture` is causing my OV5640 to heat up unnecessarily, causing my camera module to overheat on hot days (over 36°C under direct sunlight). After taking out the 5s `WaitBeforePicture`, my OV5640 hasn't overheated yet even on hot days.

I'm pretty happy with the stability of the set up now. @SybexX @Slider0007 Question is, how do we want to proceed wrt the issues below:
1. 10 MHz `xclk_freq_hz`: Do we want to make this the default and let users set a different value if it doesn't work for them?
2. `WaitBeforePicture`: Do we really need this? What problem does this actually solve? Sorry I don't know the history of it.

SybexX:
>WaitBeforePicture: Do we really need this? What problem does this actually solve? Sorry I don't know the history of it.

It is necessary because some digital meters need some time to wake up from standby mode (have to be illuminated with a light source for a certain time).

>10 MHz xclk_freq_hz: Do we want to make this the default and let users set a different value if it doesn't work for them?

I asked @jomjol and @caco3 because of that.
Since we want to publish a releas soon(next or after next week), no further changes will be carried out on the main, so this is done if after the next release.

Teleftaios:
Hi. I tried the stable 16.0.0 but autofocus (with manual soldered 3,3V) does not work. Even is there an chance to make the livestream in full resolution instead of 640?

jasaw:
> Hi. I tried the stable 16.0.0 but autofocus (with manual soldered 3,3V) does not work. Even is there an chance to make the livestream in full resolution instead of 640?

@Teleftaios autofocus support has not been released yet, so 16.0.0 does not support autofocus. If you want to try autofocus, you'll have to compile the firmware from this branch with the `xclk_freq_hz` changed to 10MHz.
I'm still working on merging this PR, but I've been busy recently.

@SybexX I've been testing this PR with 10MHz `xclk_freq_hz` and zero seconds `WaitBeforePicture` and looping image capture 10 times before the actual capture. It has been running stable on both my ov5640 with uptime of 28 days. The outstanding questions are:
1. Regarding the `WaitBeforePicture`, I think we should turn on the LED, wait for `WaitBeforePicture` 5 seconds, then only take the camera out of deep sleep. That should solve the camera module heating up too much.
2. Should we change the default value of `WaitBeforePicture`?
3. We should keep the looping image capture 10 times before the actual capture. I find that this produces stable image consistently.
4. Regarding the 10MHz `xclk_freq_hz`, should we just change the default `xclk_freq_hz` to 10MHz?
5. Should we allow users to set a new `xclk_freq_hz` value?

allexoK:
Very cool, thanks for implementing it! I will add auto-focus power to Ai-On-The-Edge-Cam boards next revision (https://github.com/jomjol/AI-on-the-edge-device/discussions/2963).

PR #3538, opened 2025-02-04
add .gz

SybexX:
The ESP32 checks whether the file requested by the web server is available on the SD as .gz, 
if this is the case, it will be sent it to the web server, otherwise the "normal file" will be sent. 
According to information from the Internet, .gz is supported by every available browser. 
Since a .gz is significantly smaller, the loading of the page is clearly accelerated, 
it is particularly noticeable if you have a bad WiFi reception or open the Data-Graph.


caco3:
@SybexX  I will do a follow-up PR and also compress the other HTML files in the pipeline.

PR #3565, opened 2025-02-16
Renamed enable/disable texts to enabled/disabled to make it more intuitive

caco3:


caco3:
@SybexX Please also use the new texts if you add new parameters.

PR #3582, opened 2025-02-24
Gz html files fix missing file server.html

caco3:


caco3:
@SybexX Thanks for the fix

PR #3590, opened 2025-02-27
Calculate md5 on upload

caco3:
The ESP calculates the MD5 sum of the uploaded OTA update file and returns it in the upload response. The UI does the same and the compares the two values. The installation is only continued if they match.

---
## New progress status:
Screenshot_20250228_222840

---

## Behavior on mismatch:
Screenshot_20250228_222848


caco3:
@SybexX I implemented now both the frontend and backen.
Can you test it on your side and merge if you are happy?

PR #3684, opened 2025-04-03
new model dig-class100-0180-s2-q

haverland:
Model updated to Tensorflow 2.17
new images, now 24300


caco3:
I think we shouldn't delete the 173 model, unless we have a reason for it!

haverland:
> I think we shouldn't delete the 173 model, unless we have a reason for it!

It will not removed on sd_card if you have an older model from older update installed. But if you think, it's fine.

PR #3697, opened 2025-04-09
Fix webinstaller update

caco3:
See https://github.com/orgs/community/discussions/154072#discussioncomment-12594580


caco3:
@jomjol FYI, it should work now as expected, see https://github.com/caco3/AI-on-the-edge-device/actions/runs/14367436108

PR #3836, opened 2025-06-23
Esp32s3 test

allexoK:
- Changed GPIO47 to GPIO12 for LEDs in HTMLs - 2.0.0 Hw version uses GPIO12 for LEDs and GPIO47 as one of the Camera pins.
- Add 'sleep while idle' option to the main code and HTMLs - makes the device sleep between rounds


allexoK:
The early board version (the green one) used GPIO12 for the camera and GPIO47 for the LEDs, making it incompatible with the ESP32-CAM standard template, which uses GPIO12 for the smart LED. I changed the config template during testing ([commit 9370e60d331da6670cbb1dcd1a58b60b77bea903](https://github.com/jomjol/AI-on-the-edge-device/pull/3836/commits/9370e60d331da6670cbb1dcd1a58b60b77bea903)

The final board version also uses GPIO12 for the smart LED, so I reverted the template file ([commit d3650e188d0789b779761c9d7e0f99b743e952df](https://github.com/jomjol/AI-on-the-edge-device/pull/3836/commits/d3650e188d0789b779761c9d7e0f99b743e952df)

So the answer to your question is:
There should be no impact on existing devices, because the pin change was reverted in a later commit.

PR #3868, opened 2025-07-24
Add option for `gal/min` rate for Home Assistant MQTT Autodiscovery

mekaneck:
This adds an option to the Home Assistant MQTT Autodiscovery meter types for `value: gal` with `rate: gal/min`. This is needed because `gal/h` is an unsupported rate, and with HA version 2025.7 (PR https://github.com/home-assistant/core/pull/146722) the sensor will no longer be created if the units are invalid for the device class used.


caco3:
Finally there is a new pre-release: https://github.com/jomjol/AI-on-the-edge-device/releases/tag/v16.1.0-RC1
Please test it and let us know within the next few weeks, so we can create the next release.

mekaneck:
Tested and confirmed working. Option is shown in the UI and, when chosen, results in the appropriate gal/min rate sensor being auto-discovered in HA.

Issue #3932, opened 2025-10-03
MQTT discovery configuration for HA deprecated in HA Core 2026.4


Kaibob2:
### The Problem

Seit HA 2025.10 erscheint folgender Eintrag im Protokoll:

```
Logger: homeassistant.components.mqtt.entity
Quelle: components/mqtt/entity.py:1455
Integration: MQTT ([Dokumentation](https://www.home-assistant.io/integrations/mqtt), [Probleme](https://github.com/home-assistant/core/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+mqtt%22))
Erstmals aufgetreten: 09:56:58 (19 Vorkommnisse)
Zuletzt protokolliert: 09:56:59

- The configuration for entity sensor.watermeter_error uses the deprecated option `object_id` to set the default entity id. Replace the `"object_id": "watermeter_error"` option with `"default_entity_id": "sensor.watermeter_error"` in your published discovery configuration to fix this issue, or contact the maintainer of the integration that published this config to fix this. This will stop working in Home Assistant Core 2026.4
- The configuration for entity sensor.watermeter_rate_per_time_unit uses the deprecated option `object_id` to set the default entity id. Replace the `"object_id": "watermeter_rate_per_time_unit"` option with `"default_entity_id": "sensor.watermeter_rate_per_time_unit"` in your published discovery configuration to fix this issue, or contact the maintainer of the integration that published this config to fix this. This will stop working in Home Assistant Core 2026.4
- The configuration for entity sensor.watermeter_rate_per_digitization_round uses the deprecated option `object_id` to set the default entity id. Replace the `"object_id": "watermeter_rate_per_digitization_round"` option with `"default_entity_id": "sensor.watermeter_rate_per_digitization_round"` in your published discovery configuration to fix this issue, or contact the maintainer of the integration that published this config to fix this. This will stop working in Home Assistant Core 2026.4
-  The configuration for entity sensor.watermeter_timestamp uses the deprecated option `object_id` to set the default entity id. Replace the `"object_id": "watermeter_timestamp"` option with `"default_entity_id": "sensor.watermeter_timestamp"` in your published discovery configuration to fix this issue, or contact the maintainer of the integration that published this config to fix this. This will stop working in Home Assistant Core 2026.4
[...]

PR #3970, opened 2025-11-25
Add `default_entity_id` to MQTT HA autodiscovery payload

mekaneck:
Add `default_entity_id` to the HA autodiscovery payload to address a deprecation (warning starting with HA 2025.11 and causing an error starting with HA 2026.4).

Addresses issue #3932 

If desired, the `object_id` (line 107) can be removed after April 2026.

### DIFF
--- a/code/components/jomjol_mqtt/server_mqtt.cpp
+++ b/code/components/jomjol_mqtt/server_mqtt.cpp
@@ -72,16 +72,16 @@ bool sendHomeAssistantDiscoveryTopic(std::string group, std::string field,
     std::string topicFull;
     std::string configTopic;
     std::string payload;
-	std::string component;
+    std::string component;
 
     configTopic = field;
 
     if (group != "" && (*NUMBERS).size() > 1) { // There is more than one meter, prepend the group so we can differentiate them
         configTopic = group + "_" + field;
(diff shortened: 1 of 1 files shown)


caco3:
Finally there is a new pre-release: https://github.com/jomjol/AI-on-the-edge-device/releases/tag/v16.1.0-RC1
Please test it and let us know within the next few weeks, so we can create the next release.

mekaneck:
Tested and confirmed working in v16.1.0-RC1. `default_entity_id` property, with the correct domain (`sensor`, `binary_sensor`, or `button`) is added to all HA MQTT autodiscovery topics.

gotsila:
Hello, 
the default_entity_id property is now right there.
seen in MQTT-Explorer. retain all in MQTT...
However, the old object_id is still being sent, which is why HA continues to report an error.

April isn't far off now. Couldn't this be removed already? As long as this message keeps appearing, I'm not sure if I've done everything correctly.

mekaneck:
Sending both the `object_id` and the `default_entity_id` is intended behavior and will not result in errors. This is done to support people who are running HA versions older than 2025.11. Other projects (like zigbee2mqtt) do this as well.

There is something else wrong with your setup.

PR #3987, opened 2025-12-09
Update WLAN configuration with new SSID and password

hampibart:


bvweerd:
Are you putting your own wifi SSID and password out on the internet right now? This seems a bit of an odd PR...

PR #4000, opened 2025-12-29
feat: Add reed contact support for intermediate meter readings

### LINKED ISSUE #3999: Add support for reed contacts
### Reed contact support

A lot of meters (gas, water, ...) have a magnet moving with the last digit and most gas meters even have a slot for a reed contact. With these meters the magnet passes by the reed contact slot with each full revolution of the last digit.
It would be great if there was a simple option to add such a reed contact to a GPIO for intermediary readings between picture analysis. 
E.g. picture analysis happens every five minutes but the last digit currently as an example revolves in 15 seconds. The reed contact would be closed every 15 seconds and the second to last digit could internally be increased and the new value published to home assistant or somewhere else. 
My specific gas meter has the magnet in the 6 on the third decimal digit. Since the slot for the reed contact is at the bottom the contact is closed from x.xx9 to x.xx1 including the 0 crossing. This may be different for other meters. 

Using only the reed contact is often not 100% reliable since sometimes a single revolution causes two triggers due to bounce or similar. This solution would use the reed contact for a very cheap to compute and rather high resolution value and use the camera more or less only for auto correction of drift due to missed reed contact closure or double closure.

I imagine the following configuration parameters for this: 
* enable reed contact [yes/no] (default: no)
* GPIO [number and input/input-pullup] (default: input-pullup)
* increase on reed close or on open [close/open] (default: open)
* digit to increase on reed close/open [select a digit] (default second to last digit if easily possible to implement)
* debounce time (e.g. 500ms) values from 0ms to 10000ms should be supported [number input in milliseconds] (default 500ms)
* publish on value increment [yes/no] (default: yes)

Xyolyp:
## Info
This is my very first time working with cpp so please do a proper review and please provide feedback and ask questions since I may misunderstood things entirely. 

Implements reed contact (magnetic switch) support for GPIO pins to enable intermediate meter readings between image analysis cycles.

## Features
- Reed contact configuration per GPIO (enable, trigger type, debounce, digit index)
- Configurable debounce time (0-10000ms) to prevent false triggers
- Trigger types: on-close or on-open
- Automatic digit value increment when reed contact triggers
- MQTT publication support for real-time updates
- Integration with image-based drift correction

## Configuration Example
```ini
[GPIO]
IO12 = input-pullup rising-and-falling 10 false false GAS_METER
ReedContactIO12 = true on-close 500 1 true
```

## Supported GPIO Pins
Pins 0, 1, 3, 4, 12, 13

## Documentation
See REED_CONTACT_SUPPORT.md for detailed documentation and troubleshooting.

Closes #3999

### DIFF
--- a/sd-card/html/edit_config_template.html
+++ b/sd-card/html/edit_config_template.html
@@ -1645,6 +1645,60 @@
             <td><input type="text" id="GPIO_IO1_value6"></td>
             <td></td>
         </tr>
+
+        <!-- Reed Contact Configuration for GPIO1 -->
+        <tr class="GPIO_IO1 GPIO_item expert">
+            <td class="indent2">
+                <span class="GPIO_IO1 GPIO_item">GPIO1 Reed Contact Enable</span>
+            </td>
+            <td><input type="checkbox" id="GPIO_IO1_ReedEnable"></td>
+            <td></td>
+        </tr>
+
+        <tr class="GPIO_IO1 GPIO_item expert">
+            <td class="indent2">
+                <span class="GPIO_IO1 GPIO_item">GPIO1 Reed Contact Trigger</span>
+            </td>
+            <td>
+                <select id="GPIO_IO1_ReedTrigger">
+                    <option value="on-close">on close</option>
+                    <option value="on-open">on open</option>
+                </select>
+            </td>
+            <td></td>
+        </tr>
+
+        <tr class="GPIO_IO1 GPIO_item expert">
+            <td class="indent2">
+                <span class="GPIO_IO1 GPIO_item">GPIO1 Reed Contact Debounce (ms)</span>
+            </td>
+            <td>
+                <input required type="number" id="GPIO_IO1_ReedDebounce" min="0" max="10000" value="500"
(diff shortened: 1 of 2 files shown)


code/components/jomjol_flowcontroll/ClassFlowPostProcessing.cpp
```diff
@@ -146,6 +146,50 @@ bool ClassFlowPostProcessing::SetPreValue(double _newvalue, string _numbers, boo
...
+
+    NumberPost* number = NUMBERS[digitIndex];
+    if (number == NULL) {
+        LogFile.WriteToFile(ESP_LOG_WARN, TAG, "incrementValueByReedContact: NumberPost at index %d is NULL", digitIndex);
+        return false;
+    }
+
+    // Get the current value and calculate the next value (increment by the smallest unit for this digit)
+    double currentValue = number->PreValue;
+    double increment = 1.0 / pow(10, number->DecimalShift + number->Nachkomma);
```
Xyolyp: Is this correct?

PR #4026, opened 2026-01-16
Esp32s3 test

allexoK:


friedpa (commented):
@allexoK: Everything works fine. I have also your LoraWAN module running (with implemented Mesghtastic Network via I2C Bus), I am not getting it right to alsoget that board in enerysaving mode, could you please give me a hint where to start!? Thanks!

PR #4122, opened 2026-06-03
Update wlan.ini

yoth78:


SybexX:
@yoth78 You have to adapt the wlan.ini on your PC and not directly in the firmware.

https://jomjol.github.io/AI-on-the-edge-device-docs/Installation/#2-firmware

Issue #4112, opened 2026-05-13
Premature digit transition accepted on early-transition meters; no history-based logic to detect impossible digit/analog combinations


Apfelstruhdl:
## Summary

On a water meter with physical "early transition" behavior (the last digit visually rotates ahead of the corresponding analog pointer wrap-around), the firmware accepts a digit increment that the analog pointer history makes physically impossible. Once accepted, the inflated value becomes the new PreValue. With `AllowNegativeRates = true`, the system eventually self-corrects via further accepted backward jumps, producing a non-monotonic published series and multiple visible error windows. With `AllowNegativeRates = false`, the wrong PreValue would persist until the meter physically catches up — potentially weeks on a low-flow meter.

Single-frame correction logic (`AnalogToDigitTransitionStart`, `CheckDigitIncreaseConsistency`) cannot resolve this because the meter face is *correctly* showing the premature digit value. Distinguishing a real rollover from a premature display state requires history of the analog pointer trajectory across previous readings.

## Environment

- Firmware: **v16.1.0** (Commit `a1ccda2`)
- Digit model: `dig-class11_1910_s2_q.tflite`
- Analog model: `ana-cont_1500_s2_q.tflite`
- 5 digits + 4 analog pointers
- `DecimalShift = 3`
- `AnalogToDigitTransitionStart = 8.5` (set low based on observed early-transition behavior)
- `MaxRateValue = 80`, `MaxRateType = RateChange`
- `AllowNegativeRates = true`
- `CheckDigitIncreaseConsistency = false`
- `PreValueUse = true`

Full config attached.

## Meter behavior

The meter is an early-transition design with a wide transition band: `dig5` (last digit before the analog pointers) begins rotating well before `ana1` reaches 9. In the attached dataset:

- `dig5` ambiguity (CNN output `N`) is first observed at `ana1 = .87` — raw `0207N888.63` at `2026-05-12 21:14 local`
- `dig5` commits to the incremented value while `ana1` is still reading `.98` — raw `02074986.18` at `21:13 UTC` / `23:13 local`

The meter face is physically showing this — the CNN is reading what's actually visible. But the *cumulative true value* of the meter is still in the `N.xxxx` range until `ana1` actually wraps through zero, somewhere in the range `ana1 = .87` to `.99` depending on the specific transition.

## Observed failure

Dataset attached (Home Assistant export, ~6 hours starting `2026-05-12 19:00 UTC`).

[...]

PR #4137, opened 2026-07-07
Add opt-in history-anchored post-processing reconciliation for premature dial transitions (#4112)

Apfelstruhdl:
## Summary

Adds an **opt-in** history-anchored reconciliation layer to post-processing that fixes a class of wrong readings on meters with "early transition" analog dials, addressing #4112.

On such meters a pointer (or digit) flips ahead of the real rollover, so a single frame can be recognised as a value that is off by one dial step (e.g. ±100 L). The existing single-frame logic and the `MaxRateValue` / `AllowNegativeRates` guards can't distinguish such a premature — and often *stable* — misread from real flow, so it gets accepted and poisons `PreValue`; a tightened rate cap then rejects the legitimate catch-up and pins the reading. Config alone can't resolve this (the flip magnitude overlaps real flow; `Analog_error` is a compile-time constant).

Default-off, so existing installs are byte-for-byte unaffected.

## How it works

A new pure, dependency-free module (`MeterReconcile`, `code/components/jomjol_flowcontroll/MeterReconcile.{h,cpp}`) reconciles each frame's recognised value against the last trusted value using the meter's physical properties:

- **Carry gate.** A value crossing a most-significant-analog-dial step (a *carry*) is accepted only with evidence that the dial below actually wrapped 9→0, as a single forward step. This is enforced on **both** "clean" and ambiguous frames, because a pointer misread part of a step off its true position need not land near an integer — so a premature carry can otherwise masquerade as a trustworthy reading. Wrap evidence comes from three independent channels:
  1. the dial-below transition observed directly since the previous frame;
  2. a short "sticky" window, since the wrap and the valid carry can land on different frames (the wrap frame may itself be a misread);
  3. **inference from the value** — a genuine carry passes the sub-position (the part below one dial step) through zero, while a premature flip leaves it unchanged. This alone recovers a stalled reading with no observed wrap and no clean frame.
- **Within-step motion** is trusted on a clean frame; on an ambiguous frame forward motion is allowed while only sub-noise backward jitter is tolerated.
- **Stuck-recovery.** After the value has been held for a sustained, stable stretch, a clean frame re-anchors across a carry (heals a wrong anchor / absorbs catch-up).
- **Gross-misread ceiling.** A physically impossible single-frame jump is never accepted.

All thresholds (`msbStep`, `noiseTol`, `maxJump`) are derived per-frame from the meter's decimal scaling, so it works for digit+analog and analog-only layouts, extended resolution or not. It complements `CheckDigitIncreaseConsistency` (which handles the digit chain) by adding the equivalent protection on the analog dial chain, and replaces the fixed rate guards only while enabled.

## Configuration

Two new per-`<NUMBER>` parameters (docs under `param-docs/`):

- `HistoryReconcile` (default `false`) — enable it.
- `HistoryMaxJump` (default: auto-derived from the dial scale) — optional gross-misread ceiling override.

No effect on sequences without analog ROIs.

## Testing

- **Host unit tests** (`code/test/host/`, 53 checks) cover the carry gate (all three evidence channels), premature-flip rejection on clean and ambiguous frames, backward/multi-step carry rejection, gross-misread rejection, scale invariance, floating-point boundary safety, and stability-bounded recovery. A CI job compiles and runs them on every push (can be dropped if you'd prefer not to touch the workflow).
- **Firmware build + packaging** pass in CI.
- **Validated on a real early-transition meter for ~1 week**, including the exact failures from #4112:
  - premature ±100 L dial flips — **held**, no longer poison `PreValue`;
  - the ±900 L 0/9-seam flicker (`ana1` read as 9.x while just past 0) — **rejected** every frame while genuine flow between them is still tracked;
  - a full multi-dial cascade rollover (`ana1` 9→0 + thousands digit + `dig5`=`N`) — **crossed cleanly in real time, zero hold frames** (the case that previously stalled for ~50 min);
  - **zero accepted poison** over 3 continuous days; running total stays exact and HA long-term statistics stay clean.

## Notes

- Opt-in + default-off ⇒ no regression risk for existing users.
- Scoped to sequences **with analog ROIs** by design: the digit chain already has its
  history-anchored guard (`CheckDigitIncreaseConsistency`), and the reconciler's evidence signals
  (boundary-band cleanliness, wrap observation, sub-position inference) are properties of
  continuously rotating pointers — a digit wheel parks *at* integers, so the ambiguity model does
  not transfer. Extending it to digit-only meters with continuous digit models would be possible
  future work, but needs its own ambiguity model and validation.
- Operating assumption (documented): typical consumption between two cycles stays well below one
  most-significant-dial step. With very long capture intervals, heavy flow can exceed that
  per-cycle step; the value then holds during the burst and catches up shortly after flow pauses
  (nothing lost, but readings lag). A time-aware accept path for such multi-step catch-up would be
  possible future work.
- Recognition scatter is the root trigger; reducing it further (e.g. enabling `AntiAliasing` for the alignment rotation) lowers how often the guard has to engage — measured ~50 % less analog scatter on the test meter — but that's independent of this change.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -109,6 +109,19 @@ jobs:
         cp -r ./sd-card/demo/* ./demo/
 
 #########################################################################################
+## Host unit tests (post-processing reconciliation logic)
+#########################################################################################
+  test-reconcile:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build & run reconciliation host unit tests
+      run: |
+        cd code/test/host
+        g++ -std=c++11 -Wall -Wextra -I ../../components/jomjol_flowcontroll test_meter_reconcile_host.cpp ../../components/jomjol_flowcontroll/MeterReconcile.cpp -o reconcile_test
+        ./reconcile_test
+
+#########################################################################################
 ## Pack for Update
 #########################################################################################
   pack-for-update:
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ### Core Changes and Bug fixes
 - Added opt-in history-anchored post-processing reconciliation (`HistoryReconcile`, with `HistoryMaxJump`) that fixes premature analog-dial transitions (single-frame ±1 dial flips) and values getting stuck at digit rollovers, by trusting the reading on unambiguous frames and gating a most-significant analog-dial carry on an observed wrap of the dial below it [#4112](https://github.com/jomjol/AI-on-the-edge-device/issues/4112)
+- `HistoryReconcile` is now preserved by the config editor (previously silently dropped on every ROI/alignment save) and is directly editable in the expert config page
 
 
 # [16.1.0] - 2026-01-11
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+# [Unreleased]
+
+### Core Changes and Bug fixes
+- Added opt-in history-anchored post-processing reconciliation (`HistoryReconcile`, with `HistoryMaxJump`) that fixes premature analog-dial transitions (single-frame ±1 dial flips) and values getting stuck at digit rollovers, by trusting the reading on unambiguous frames and gating a most-significant analog-dial carry on an observed wrap of the dial below it [#4112](https://github.com/jomjol/AI-on-the-edge-device/issues/4112)
+
+
 # [16.1.0] - 2026-01-11
 
 For a full list of changes see [Full list of changes](https://github.com/jomjol/AI-on-the-edge-device/compare/v16.0.0...v16.1.0)
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ### Core Changes and Bug fixes
 - Added opt-in history-anchored post-processing reconciliation (`HistoryReconcile`, with `HistoryMaxJump`) that fixes premature analog-dial transitions (single-frame ±1 dial flips) and values getting stuck at digit rollovers, by trusting the reading on unambiguous frames and gating a most-significant analog-dial carry on an observed wrap of the dial below it [#4112](https://github.com/jomjol/AI-on-the-edge-device/issues/4112)
 - `HistoryReconcile` is now preserved by the config editor (previously silently dropped on every ROI/alignment save) and is directly editable in the expert config page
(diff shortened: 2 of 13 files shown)


Apfelstruhdl:
Added a 4th commit: registers HistoryReconcile in the config editor and adds a GUI toggle. Without it, the parameter was silently dropped from config.ini on every ROI/alignment save (the editor rebuilds the file from a fixed key whitelist), so it couldn't be used reliably. Also now been running on real hardware for a few weeks.

Apfelstruhdl:
Update after ~2 months running on a live meter.

Correction to the description above: the host test suite is now 58 checks, not 53 — the additional ones cover the change below.

A fifth commit was added: Bound the reconciliation stuck-recovery to a single dial step. It came out of a real failure, which seems worth documenting because it shows where the original design was too permissive.

The recovery path accepted a carry of any size once the value had been held for recoverHolds cycles on a clean frame with a stable recognition. "Clean and stable" turns out not to be evidence that a far-away reading is correct — a digit that reads one position ahead sits perfectly clean and perfectly stable for as long as the meter is idle. On 10 Aug my meter's last m³ digit transitioned early; the reconciler correctly held the misread for ~10 hours, then recovery accepted it, leaving the published value exactly +1000 L high. It stayed that way (raw correct, value poisoned, so every later frame read backwards and was held) until I set the pre-value by hand five days later.

The fix caps recovery at recoverMaxSteps (default 1) most-significant-dial steps. One step is all recovery needs, since its purpose is healing an anchor left behind by a missed carry. A reading further out keeps being held — which is visible (the value stops advancing and diverges from raw) and leaves the totals intact — and is then resolved by the normal PreValueAgeStartup re-seed or a manual pre-value. Note this is independent of MaxRateValue: that ceiling is derived from the dial scale and is far too wide to catch a single-digit misread.

Field evidence, in both directions:

The gate working as intended: on 28 Aug the raw reading dropped by 4000 L (an m³ digit misread) and stayed wrong for ~6 hours. The value correctly froze rather than following it down, then resumed cleanly once raw recovered — divergence back to exactly 0.00, no corruption, negligible consumption lost.
Why the cap matters: between 1–10 Sep, on firmware that had the reconciler but not the cap, five separate early-digit misreads were accepted through the unbounded recovery path, totalling ~11,000 L. Measured against the cap afterwards, those jumps were 9, 13 and 88 dial steps — all would have been held.

Still opt-in and default-off, so no behaviour change for existing users.

PR #4150, opened 2026-08-13
Retry camera initialization after power reset

Friendbg:
## Summary

This change retries `esp_camera_init()` up to 5 times before giving up.

Some ESP32-CAM boards occasionally fail to initialize the camera after a reboot, returning errors such as `0x00000004`. In practice, another attempt after a short delay often succeeds.

## Changes

- retry `esp_camera_init()` up to 5 times
- 500 ms delay between retries
- improved logging showing the attempt number
- preserve existing behaviour if all retries fail

## Testing

Tested on two ESP32-CAM boards.

On one board the camera previously failed to initialize after the first reboot. After this change I performed multiple consecutive restarts without reproducing the issue.

No regressions observed.


SybexX:
The current main should [fix](https://github.com/jomjol/AI-on-the-edge-device/pull/4127) the problem with 0x00000004 for most users.
The problem with 0x00000004 has been known for a long time and has different causes depending on the ESP32 cam hardware.

possible problems:
- The supply voltage drops at the Boot
- xclk_freq_mhz = 20 is too high for the camera and must be reduced to 8
- and many others.............

https://github.com/jomjol/AI-on-the-edge-device/issues?q=is%3Aissue%20state%3Aopen%200x00000004
https://github.com/jomjol/AI-on-the-edge-device/discussions?discussions_q=0x00000004

